PR #34, opened 2015-02-21
Support package names in the module comments. Fixes #17.

### LINKED ISSUE #17: Package names
How does *fades* the name of the package that it should install?

I think it's taking the name from the `import` line but sometimes this is not correct: for example: beautifulsoup.

I have this in my `.py` file:
```
from bs4 import BeautifulSoup  # fades.pypi
```

and *fades* tries to install the package named `bs4` which is incorrect because the proper name is: `beautifulsoup4`

I suggest to use something like this:
```
from bs4 import BeautifulSoup  # fades.pypi beautifulsoup4
```

facundobatista:
Code and tests, not only adding the functionality for "projectname" itself, but also exploding on bad comparisons like "<>".

Note: when we search for a comment, before we find "fades.pypi", we just log in warning, but after that the syntax is rigid, so if the user fails to write that properly and we don't understand it, it's better to raise an Exception.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -74,6 +74,13 @@ Also, you can indicate a particular version condition, examples::
     import somemodule   # fades.pypi == 3
     import somemodule   # fades.pypi >= 2.1
 
+Sometimes, the project itself doesn't match the name of the module; in
+these cases you can specify the project name (optionally, before the
+version)::
+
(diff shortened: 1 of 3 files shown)


gilgamezh:
Great!

PR #46, opened 2015-02-26
New tests

gilgamezh:
Some new tests and I fixed one test that was crashing.


gilgamezh:
@facundobatista #50 fixed here :)

PR #85, opened 2015-06-06
Different interpreter versions

### LINKED ISSUE #6: Support different Python versions for child execution
This will be automatically decided by fades being executed like 'fades' or 'fades3', or 'fades2.7', etc...

gilgamezh:
:foggy: :grin:
resolve #6


fades/helpers.py
```diff
@@ -48,3 +50,46 @@ def get_basedir():
         logger.debug("Package xdg not installed; using ~/.fades folder")
         from os.path import expanduser
         return expanduser("~/.fades")
+
+
+def _get_interpreter_info(interpreter=None):
+    """Returns the interpreter's full path using pythonX.Y format and his version"""
```
facundobatista: Be imperative ("Return", no "Returns"), and finish the sentence with a "."

fades/helpers.py
```diff
@@ -48,3 +50,46 @@ def get_basedir():
...
+    if interpreter is None:
+        # If interpreter is None by default returns the current interpreter data.
+        major, minor = sys.version_info[:2]
+        executable = sys.executable
+    else:
+        cmd = ("import sys, json; d = dict(path=sys.executable);"
```
facundobatista: Build this cmd as a module-level constant, in a multiline string for clarity, like:

SHOW_VERSION_CMD = """
import sys, json
d = dict(path=sys.executable)
d.update(zip('major minor micro releaselevel serial'.split(), sys.version_info))
print(json.dumps(d))
"""

fades/helpers.py
```diff
@@ -48,3 +50,46 @@ def get_basedir():
...
+        minor = requested_interpreter_info['minor']
+    if executable[-1].isdigit():
+        executable = executable.split(".")[0][:-1]
+    interpreter = "{}{}.{}".format(executable, major, minor)
+    interpreter_version = "{}.{}".format(major, minor)
+    return interpreter, interpreter_version
+
+
+def get_interpreter_version(requested_interpreter):
+    """ Return a 'sanitized' interpreter, its version and indicates if it is the current one. """
```
facundobatista: Don't leave those spaces around the sentence in the docstring

fades/helpers.py
```diff
@@ -48,3 +50,46 @@ def get_basedir():
...
+    logger.debug('Getting interpreter version for: %s', requested_interpreter)
+    current_interpreter, current_version = _get_interpreter_info()
+    logger.debug('Current interpreter version is %s', current_version)
+    requested_interpreter, requested_version = _get_interpreter_info(requested_interpreter)
```
facundobatista: if requested_interpreter is None (I expect that to be the huge majority of cases) this second _get_interpreter_info call is unnecessary...

facundobatista:
Other general comment: if venvscache is using interpreter_version, that means that it will mix venvs for "/usr/bin/python2.7" and "/usr/local/bin/python2.7".

Do we want that? Why don't just use the "interpreter" (which is the full path interpreter with version and everything, right?)

gilgamezh:
I think I produced a problem trying to solve another.
As it is now I was trying to store only the version, thinking of *"givme a venv for x.y version" and not for a particular one"*. After that we decided to store only major and minor version. But I realize now that doing it in that way doesn't make sense to identify the 'interpreter'.

If we store the 'interpreter' we will have a unique venv for each 'interpreter'.

So, do we want to ask fades for a '*virtualenv with the same version that --python*" or for *"a virtualenv with the same interpreter that  --python*"?

The last option could generate more virtualenvs in the future but fits better for rare use cases. I prefer it

What do you think?

facundobatista:
+1 to do it with fullpath and major/minor versions... example: "/usr/bin/python2.7"

gilgamezh:
@facundobatista it's your turn

facundobatista:
Final fix! It's "--no-pip", not "--nopip".

Thanks,

gilgamezh:
Done!

facundobatista:
Thanks!!!

PR #99, opened 2015-08-15
Make pip install pip where we need to and not where it wrongly assumes when it's inside a venv.

facundobatista:


gilgamezh:
inception!! :+1:

PR #101, opened 2015-08-16
added an option to execute ipython

arielrossanigo:
Added an option use ipython instead of python shell. 

The option is -i | --ipython. This option add ipython as a dependency.

Examples:
`$ fades -i -d requests`

`$ echo "import requests; print(requests.__version__)" > foo.py`
`$ fades -i -d requests foo.py -i`

In the last example, first *i* means "use IPython" while second *i* is a parameter to IPython shell indicating that we want to use interactive session.

### DIFF
--- a/fades/main.py
+++ b/fades/main.py
@@ -82,6 +82,8 @@ def go(version, argv):
     parser.add_argument('-p', '--python', action='store',
                         help=("Specify the Python interpreter to use.\n"
                               " Default is: %s") % (sys.executable,))
+    parser.add_argument('-i', '--ipython', action='store_true',
+                        help="use IPython shell.")
     parser.add_argument('child_program', nargs='?', default=None)
     parser.add_argument('child_options', nargs=argparse.REMAINDER)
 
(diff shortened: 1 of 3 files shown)


fades/main.py
```diff
@@ -82,6 +82,8 @@ def go(version, argv):
     parser.add_argument('-p', '--python', action='store',
                         help=("Specify the Python interpreter to use.\n"
                               " Default is: %s") % (sys.executable,))
+    parser.add_argument('-i', '--ipython', action='store_true',
+                        help="use IPython shell.")
```
facundobatista: Please add -i/--ipython options and explanation both to README and man page.

fades/main.py
```diff
@@ -119,13 +121,19 @@ def go(version, argv):
         l.warning("Overriding 'quiet' option ('verbose' also requested)")
 
     # parse file and get deps
+    if args.ipython:
+        ipython_dep = parsing.parse_manual(['ipython'])
```
facundobatista: Please log in debug that you're adding the ipython dependency

fades/main.py
```diff
@@ -119,13 +121,19 @@ def go(version, argv):
...
+    else:
+        ipython_dep = []
     indicated_deps = parsing.parse_srcfile(args.child_program)
     l.debug("Dependencies from source file: %s", indicated_deps)
     reqfile_deps = parsing.parse_reqfile(args.requirement)
     l.debug("Dependencies from requirements file: %s", reqfile_deps)
     manual_deps = parsing.parse_manual(args.dependency)
     l.debug("Dependencies from parameters: %s", manual_deps)
-    indicated_deps = _merge_deps(indicated_deps, reqfile_deps, manual_deps)
+    indicated_deps = _merge_deps(
```
facundobatista: No need to have multiple lines here, as column limit is 99 in this project :)

facundobatista:
Awesome branch, thanks!! I added some inline comments ↑

arielrossanigo:
I've just add some commits in order to support your comments

facundobatista:
Awesome, thank you very much!

PR #103, opened 2015-08-16
Reqs parsed from docstrings

matuu:


gilgamezh:
:+1: 
We have to solve the merge conflicts and wait for @facundobatista +1

fades/parsing.py
```diff
@@ -106,6 +107,32 @@ def _parse_content(fh):
...
 
 
+def _parse_docstring(fh):
+    """Parse the docstrings of a script to find marked dependencies."""
+    requirements = []
+    # get docstring into source file
+    ds_parsed = ast.parse(fh.read())
+    ds_texts = [ds_parsed]
+    ds_texts.extend([node for node in ds_parsed.body if isinstance(node, ast.FunctionDef) or
+                     isinstance(node, ast.ClassDef) or isinstance(node, ast.Module)])
```
facundobatista: Do we want to get the docstrings for everything? Or just the module? I vote for "module only"

facundobatista:
Gilgamezh, Matías, I added an inline comment about somethin we need to discuss... the rest, I like it :)

fades/parsing.py
gilgamezh: :+1: module only!

matuu:
I fixed this and I opened (#111) another pull request.

PR #105, opened 2015-08-16
First attempt to use fades as a module: python -m fades 

jmansilla:


facundobatista:
Awesome, thank you very much!

PR #109, opened 2015-08-22
Remove virtualenv

ricardokirkner:
This PR introduces a new --rm flag for fades that allows it to remove the virtualenv matching the specified dependencies.

### DIFF
--- a/fades/cache.py
+++ b/fades/cache.py
@@ -26,6 +26,8 @@ from contextlib import contextmanager
 
 from pkg_resources import Distribution
 
+from fades import helpers
+
 logger = logging.getLogger(__name__)
 
 
@@ -72,22 +74,28 @@ class VEnvsCache:
         # it did it through!
         return True
 
-    def _select(self, current_venvs, requirements, interpreter):
+    def _select(self, current_venvs, requirements=None, interpreter='', uuid=''):
         """Select which venv satisfy the received requirements."""
-        logger.debug("Searching a venv for reqs: %s and interpreter: %s",
-                     requirements, interpreter)
+        if uuid:
+            logger.debug("Searching a venv by uuid: %s", uuid)
+            env_path = os.path.join(helpers.get_basedir(), uuid)
+            match = lambda env: env.get('metadata', {}).get('env_path') == env_path
+        else:
+            logger.debug("Searching a venv for reqs: %s and interpreter: %s",
+                         requirements, interpreter)
+            match = lambda env: (env.get('interpreter') == interpreter and
+                                 self._venv_match(venv['installed'], requirements))
         for venv_str in current_venvs:
             venv = json.loads(venv_str)
-            if venv.get("interpreter") == interpreter:
-                if self._venv_match(venv['installed'], requirements):
-                    logger.debug("Found a matching venv! %s", venv)
-                    return venv['metadata']
+            if match(venv):
+                logger.debug("Found a matching venv! %s", venv)
+                return venv['metadata']
         logger.debug("No matching venv found :(")
 
-    def get_venv(self, requirements, interpreter):
+    def get_venv(self, requirements=None, interpreter='', uuid=''):
         """Find a venv that serves these requirements, if any."""
         lines = self._read_cache()
-        return self._select(lines, requirements, interpreter)
+        return self._select(lines, requirements, interpreter, uuid=uuid)
 
     def store(self, installed_stuff, metadata, interpreter):
         """Store the virtualenv metadata for the indicated installed_stuff."""
--- a/fades/cache.py
+++ b/fades/cache.py
@@ -17,10 +17,12 @@
 
 """The cache manager for virtualenvs."""
 
+import fcntl
 import json
 import logging
 import os
 import time
+from contextlib import contextmanager
 
 from pkg_resources import Distribution
 
@@ -84,12 +86,7 @@ class VEnvsCache:
 
     def get_venv(self, requirements, interpreter):
         """Find a venv that serves these requirements, if any."""
-        if os.path.exists(self.filepath):
-            with open(self.filepath, 'rt', encoding='utf8') as fh:
-                lines = [x.strip() for x in fh]
-        else:
-            logger.debug("Index not found, starting empty")
-            lines = []
+        lines = self._read_cache()
         return self._select(lines, requirements, interpreter)
 
     def store(self, installed_stuff, metadata, interpreter):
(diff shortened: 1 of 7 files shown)


man/fades.1
```diff
@@ -68,6 +68,14 @@ Select which Python version to be used; the argument can be just the number (2.7
 
 The dependencies can be indicated in multiple places (in the Python source file, with a comment besides the import, in a \fIrequirements\fRfile, and/or through command line. In case of multiple definitions of the same dependency, command line overrides everything else, and requirements file overrides what is specified in the source code.
 
+.TP
+.BR -x ", " --exec
+Look up the \fIchild_program\fR in the virtualenv.
```
facundobatista: Silly change... don't like the "look up", we're *executing* it, maybe use Execute, or Run, or something that indicates better that that process is being run?

fades/main.py
```diff
@@ -179,3 +181,10 @@ def _signal_handler(signum, _):
     rc = p.wait()
     if rc:
         l.debug("Child process not finished correctly: returncode=%d", rc)
+
+    if args.remove:
+        # remove this venv from the cache
+        env_path = venv_data.get('env_path')
+        if env_path:
+            venvscache.remove(env_path)
```
facundobatista: what if env_path is empty or none? what case does it represent? maybe we should log a warning saying something like "you attempted to remove a venv that doesn't exists"?

fades/cache.py
```diff
@@ -102,5 +99,42 @@ def store(self, installed_stuff, metadata, interpreter):
...
+        """Read virtualenv metadata from cache."""
+        if os.path.exists(self.filepath):
+            with open(self.filepath, 'rt', encoding='utf8') as fh:
+                lines = [x.strip() for x in fh]
+        else:
+            logger.debug("Index not found, starting empty")
+            lines = []
+        return lines
+
+    def _write_cache(self, lines, append=False):
```
facundobatista: we never call this method with append in True, False, right? in that case this option wouldn't make sense...

facundobatista:
General comment (after adding some inline comments)...

I really like the "locking" changes.

OTOH, what is the purpose of this "--rm" behaviour? In which case(s) you would use it? Why it's needed?

Thanks!

fades/cache.py
ricardokirkner: we do call it with append=True, in the store method (l.103 of this diff)

fades/main.py
ricardokirkner: fixed

man/fades.1
ricardokirkner: fixed

ricardokirkner:
The purpose of the --rm behaviour is to clean up old virtualenvs from disk. When you're certain you don't want to work with a specific set of requirements anymore you can pass --rm to remove the virtualenv from disk. This initial use case might not be very useful, but this is also a building block for a future use case to have fades clean up virtualenvs by other constraints, for example, remove all virtualenvs older than a specific date, or ones that haven't been used in certain amount of time, etc.

fades/cache.py
facundobatista: but we never call it with append in False, or not adding the append parameter, right?

ricardokirkner: not explicitly, but the default is append=False which is the use case we want for writing the cache contents after removing one entry (we basically overwrite the whole file)

facundobatista:
I have mixed feelings with this functionality.

I'm (as I said) a big +1 for the locking changes.

And I'm +1 for the venv removals.

What I don't like is exposing this functionality like this. In the current proposal, user needs to know the venv is ephimeral before even calling fades, so to pass the --rm option.

I'm not even sure about post-cleanup (like calling "fades --rm 6d2c3d4a-903e-461a-8533-151d6268e63d").

In *my* head, this always worked as a /etc/cron.daily/ entry that called something like "fades --purge=60"  that removed venvs unused for more than 2 months.

The bottom line is that you SHOULDN'T care about created venvs. Not to clean them, not to think if they're ephimeral or not. 

Gilgamezh, what do you think?

fades/cache.py
facundobatista: Ah, right, found now where I was confused. Thanks!

ricardokirkner:
On Mon, Aug 31, 2015 at 11:27 PM, Facundo Batista <notifications@github.com>
wrote:

> I have mixed feelings with this functionality.
>
> I'm (as I said) a big +1 for the locking changes.
>
> And I'm +1 for the venv removals.
>
> What I don't like is exposing this functionality like this. In the current
> proposal, user needs to know the venv is ephimeral before even calling
> fades, so to pass the --rm option.
>

That is incorrect. The user doesn't have to know anything about the
virtualenv, except that he/she likes it to be removed. If the user executes
fades --rm whatever virtualenv is used (whether it was just created or it
already existed and was just reused) will be removed from the cache and
disk.


> I'm not even sure about post-cleanup (like calling "fades --rm
> 6d2c3d4a-903e-461a-8533-151d6268e63d").
>

One caveat. The use case is not fades --rm <hash>, but fades --rm -r
requirements.txt (ie, you never have to know the virtualenv's location).


> In *my* head, this always worked as a /etc/cron.daily/ entry that called
> something like "fades --purge=60" that removed venvs unused for more than 2
> months.
>

This particular use case comes next and builds on top of this code.


> The bottom line is that you SHOULDN'T care about created venvs. Not to
> clean them, not to think if they're ephimeral or not.
>

The user still doesn't have to think about virtualenvs.

I agree this feature is not of high value until we implement the purge
functionality like you mention in the example, but it's a stepping stone.
I think the main issue you see here is that this feature should not be used
by end users. For that one option we could do is to have an 'expert' mode
that enables features otherwise not available to users. For example

fades --help

-d ...
-r ...
-p ...
-i ...
--expert
  --rm
  --purge
  ...

facundobatista:
Mmm... I see now. So, this --rm option is to only remove it, not use it and discard.

Ok. I like it way more. It's aligned to what I want (I had prepared an idea that I wanted to write here, but after reading your response I think we're almost in the same line).

What feels weird to me is the indirection of referring to a venv. Let's agree that the "--rm" option will be an "expert" switch. I think we should request the user to pass the venv ID here.

So, NOT something like:

     fades --rm -d foo -d "bar == 0.3"

because that's a little error prone, in the sense that it needs to be kind of exact to match the venv you want to kill.

I think this is more explicit:

     fades --rm 6d2c3d4a-903e-461a-8533-151d6268e63d

And there's even less confusion about the --rm meaning...

What do you think?

ricardokirkner:
On Tue, Sep 1, 2015 at 8:49 AM, Facundo Batista <notifications@github.com>
wrote:

> Mmm... I see now. So, this --rm option is to only remove it, not use it
> and discard.
>

Well, it does remove it after running whatever you run, so it also works as
use it and discard


> Ok. I like it way more. It's aligned to what I want (I had prepared an
> idea that I wanted to write here, but after reading your response I think
> we're almost in the same line).
>
> What feels weird to me is the indirection of referring to a venv. Let's
> agree that the "--rm" option will be an "expert" switch. I think we should
> request the user to pass the venv ID here.
>
> So, NOT something like:
>
>  fades --rm -d foo -d "bar == 0.3"
>
> because that's a little error prone, in the sense that it needs to be kind
> of exact to match the venv you want to kill.
>
> I think this is more explicit:
>
>  fades --rm 6d2c3d4a-903e-461a-8533-151d6268e63d
>
> And there's even less confusion about the --rm meaning...
>
> What do you think?
>

I like it! If we combine this with other 'expert' features I'll like it
even more. For example, if we have an expert command to report which
virtualenvs are using a particular set of requirements

fades --expert --list-virtualenvs -d "bar == 0.3"

uuid1
uuid2

or, if we have one that reports virtualenvs using certain amount of disk

fades --expert --list-virtualenvs --size +100M

uuid1

Then you can do

fades --expert --rm uuid1

to remove them

facundobatista:
El 01/09/15 a las 11:13, Ricardo Kirkner escibió:

> I like it! If we combine this with other 'expert' features I'll like it
> even more. For example, if we have an expert command to report which
> virtualenvs are using a particular set of requirements
> 
> fades --expert --list-virtualenvs -d "bar == 0.3"
> 
> uuid1
> uuid2
> 
> or, if we have one that reports virtualenvs using certain amount of disk
> 
> fades --expert --list-virtualenvs --size +100M
> 
> uuid1
> 
> Then you can do
> 
> fades --expert --rm uuid1
> 
> to remove them

No need for special parameters, just call fades with -v and see it.

My problem with adding and adding parameters is that it complicates an
(already non trivial) user interface :/
-- 
.   Facundo
.
Blog: http://www.taniquetil.com.ar/plog/
PyAr: http://www.python.org.ar/
Twitter: @facundobatista

facundobatista:
I'm +1, rocanrol!!

gilgamezh:
:+1:

PR #111, opened 2015-08-27
Feature/parse req docstring

matuu:


facundobatista:
Awesome! Thank you very much!!

PR #114, opened 2015-09-07
New --system-site-packages option. 

gilgamezh:
Also --virtualenv-options and --pip options to pass custom settings to each one

Please test it.

### DIFF
--- a/fades/envbuilder.py
+++ b/fades/envbuilder.py
@@ -53,9 +53,10 @@ class FadesEnvBuilder(EnvBuilder):
 
         super().__init__(with_pip=self.pip_installed, symlinks=True)
 
-    def create_with_virtualenv(self, interpreter):
+    def create_with_virtualenv(self, interpreter, virtualenv_options):
         """Create a virtualenv using the virtualenv lib."""
         args = ['virtualenv', '--python', interpreter, self.env_path]
+        args.extend(virtualenv_options)
         if not self.with_pip:
             args.insert(3, '--no-pip')
         try:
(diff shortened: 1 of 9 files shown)


gilgamezh:
Merge conflicts solved but isn't ready to merge. 
It need more tests and update the docs.

fades/pipmanager.py
```diff
@@ -83,5 +87,5 @@ def _brute_force_install_pip(self):
 
         logger.debug("Installing PIP manually in the virtualenv")
         python_exe = os.path.join(self.env_bin_path, "python")
-        helpers.logged_exec([python_exe, self.pip_installer_fname])
+        helpers.logged_exec([python_exe, self.pip_installer_fname, '-I'])
```
facundobatista: Why this -I?

fades/envbuilder.py
```diff
@@ -72,14 +73,19 @@ def create_with_virtualenv(self, interpreter):
...
 
-    def create_env(self, interpreter, is_current):
+    def create_env(self, interpreter, is_current, options):
         """Create the virtualenv and return its info."""
         if is_current:
-            logger.debug("Creating virtualenv with pyvenv")
+            # apply pyvenv options
+            pyvenv_options = options['pyvenv_options']
+            if "--system-site-packages" in pyvenv_options:
+                self.system_site_packages = True
```
facundobatista: Shouldn't we remove this option from pyvenv_options? I mean, if we leave it, won't we "apply" it twice?

facundobatista:
A couple of details, but in general +1, because I prefer to have this only for venv and not pip, to see how it "impacts" in everything.

However: one important detail: please add docs to README and man.

Thanks!!

fades/pipmanager.py
gilgamezh: I'll add a comment. Is to always ignore installed dependencies (because we always use a new virtualenv). This is necessary when you use system-site-packages

PR #151, opened 2015-12-21
Transmit the return code from child back to whom called fades. Fixes #145.

### LINKED ISSUE #145: return the child exit code
fades is not returning the child exit code 

```
 /tmp  cat foo.py 


print("voy a fallar")
raise ValueError
 /tmp  python foo.py 
voy a fallar
  File "foo.py", line 4, in <module>
    raise ValueError
ValueError
 ✘  /tmp  echo $?
1
 /tmp  fades foo.py 
voy a fallar
  File "foo.py", line 4, in <module>
    raise ValueError
ValueError
 /tmp  echo $?     
0
```

facundobatista:


gilgamezh:
thanks! :+1:

PR #157, opened 2016-01-12
Python and the Management of Dependencies.

facundobatista:


gilgamezh:
+1 

To have a link in rtfd you have to add `pydepmanag` in the `toctree` section at index.rst.

Issue #149, opened 2015-12-19
Split out the dependencies parsing and merging from go()


facundobatista:
Just an attempt to cleanup a little that too-big main function.

Of course, what is put everywhere else, should be tested :)

andresdelfino:
HI, I'm working on this one, but my branch just adds a _get_deps to put that logic out of go(), but I saw that the _merge_deps already existed at the time this issue was opened, and the title ("Split out the dependencies parsing *and merging* from go()") got me confused when I read it a second time.

This is my branch: https://github.com/PyAr/fades/compare/master...andresdelfino:put-go-on-diet?expand=1

I have to work on the the test cases yet.

Could you confirm if I got the idea of the issue or if there's something I'm missing?

facundobatista:
Yes, that's the general idea.

However, I'd rename the new function, maybe called `consolidate_dependencies`, as you're not "getting" them from anywhere (mostly), just grouping them, but your choice.

And please improve the function argument names, more on the "what means for this function" meanings.. like `needs_ipython`, `child_program`, `requirement_files`, `manual_dependencies` (this last one mainly because "dependencies" was too generic).

The `executable` parameter it's another thing. Why do we need it? Can't we just decide for the case where `child_program` is not there? (`None` or `""`?). This is one of the things you need to improve in this branch.

Finally: probably the small `_merge_deps` can be just inside (inline) this new function...

Thanks!!!

andresdelfino:
I updated the branch with all changes applied, except the one regarding executable parameter.

I believe there's no problem if no child is specified, but if one **is** specified, we need to know if the child program is provided by the venv or not in order to parse it or not. Does this make any sense?

andresdelfino:
After a quick Telegram chat we got this things solved. I updated the branch accordingly. I was thinking about making the PR when I have the tests ready; if you think otherwise, please tell me so.

facundobatista:
Looks solid. Time for testing! 

Thanks!!!

andresdelfino:
Four things regarding the PR:

1. I changed the way child programs are dealt with in consolidate_dependencies. Instead of asking for a _is_child_from_venv parameter, we just let the caller have this logic. If no child program is specified, or if it is installed in the venv, the caller must pass None, or nothing at all, see below.

2. I added defaults for all parameters. They allow simpler calls in the test suite, and might be useful for someone using the function.

[...]

PR #169, opened 2016-02-01
Config files

gilgamezh:


README.rst
```diff
@@ -208,6 +208,28 @@ Examples:
 
 ``fades -d requests --pip-options="--index-url="http://example.com"``
 
+Setting options using config files
+----------------------------------
+
+You can also configure fades using `.ini` config files. fades will search for config files in 
+`/etc/fades.ini`, `~/.fades` and `.fades`.
```
facundobatista: it may be ~/.fades.ini or what 'xdg' indicates for your system (for example, ~/.config/fades.ini

facundobatista: Also, it's ~/.fades.ini and .fades.ini

README.rst
```diff
@@ -208,6 +208,28 @@ Examples:
...
+You have to use the same configurations that in the CLI:: 
+
+    [fades]
+    ipython=true
+    verbose=true
+    python=python3
+    dependency=requests,django>=1.8
+
+There is a little difference in how fades handle these settings: "dependecy", "pip-options" and 
+"virtualenv-options". In this cases you have to use comma separated.
```
facundobatista: s/this/these/

README.rst
```diff
@@ -208,6 +208,28 @@ Examples:
...
+
+    [fades]
+    ipython=true
+    verbose=true
+    python=python3
+    dependency=requests,django>=1.8
+
+There is a little difference in how fades handle these settings: "dependecy", "pip-options" and 
+"virtualenv-options". In this cases you have to use comma separated. 
+The most important thing is that this options will be merged. So if you configure in `~/.fades.ini`
```
facundobatista: s/this/these/

fades/file_options.py
```diff
@@ -0,0 +1,94 @@
+# Copyright 2014-2016 Facundo Batista, Nicolás Demarchi
```
facundobatista: it's a new file, starts the copyright in 2016 directly

fades/file_options.py
```diff
@@ -0,0 +1,94 @@
...
+# WITHOUT ANY WARRANTY; without even the implied warranties of
+# MERCHANTABILITY, SATISFACTORY QUALITY, or FITNESS FOR A PARTICULAR
+# PURPOSE.  See the GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License along
+# with this program.  If not, see <http://www.gnu.org/licenses/>.
+#
+# For further info, check  https://github.com/PyAr/fades
+
+"""Parse fades options from confir files."""
```
facundobatista: s/confir/config/

fades/file_options.py
```diff
@@ -0,0 +1,94 @@
...
+
+import logging
+from configparser import ConfigParser
+
+logger = logging.getLogger(__name__)
+
+CONFIG_FILES = ("/etc/fades.ini", "~/.fades.ini", ".fades")
```
facundobatista: here you need to include the proper "home config" file, from xdg, as we're doing in helpers.py for other stuff

facundobatista: also, how this plays in the Windows world? is "~" expanded correctly? what about "/etc/..."?

fades/file_options.py
```diff
@@ -0,0 +1,94 @@
...
+
+MERGEABLE_CONFIGS = ("dependency", "pip_options", "virtualenv-options")
```
facundobatista: I'd love to see a test case that grabs this variable, and then grabs the structure to parse the command line from main.py and asserts that the mergeables are the same; however, don't stuck on this test, I'm ok if you open an issue to build this test in the future and don't get stuck on this.

fades/file_options.py
```diff
@@ -0,0 +1,94 @@
...
+
+
+def get_parsers(config_files=CONFIG_FILES):
+    """
```
facundobatista: First line of text should be in this very line.

fades/file_options.py
```diff
@@ -0,0 +1,94 @@
...
+    Search for config files return a ConfigParser for each one.
+
+    Return a Configparser with all files merged in the first position and one
+    Configparser for each file ordered by priority.
+    """
+    parser = ConfigParser()
+    existing_files = parser.read(config_files, encoding='cp1250')
```
facundobatista: cp1250???

fades/file_options.py
```diff
@@ -0,0 +1,94 @@
...
+            return [parser]
+        parsers = []
+        parsers.append(parser)  # first parser is the one with all the settings.
+        for config_file in existing_files:
+            single_parser = ConfigParser()
+            single_parser.read(config_file, encoding='cp1250')
+            if parser.sections():  # avoid empty files.
+                parsers.append(single_parser)
+            return parsers
+
```
facundobatista: it looks there is a lot of redundancy here, but I'd need testcases to be sure I'm not missing a particular case...

fades/file_options.py
```diff
@@ -0,0 +1,94 @@
...
+
+def merge_parsers(parsers):
+    """Merge options in MERGEABLE_CONFIGS in a uniq ConfigParser."""
+    if len(parsers) == 1:
```
facundobatista: this "if len() == 1" can be avoided right? keep it simple

fades/file_options.py
```diff
@@ -0,0 +1,94 @@
...
+    parser = ConfigParser()
+    existing_files = parser.read(config_files, encoding='cp1250')
+    if existing_files and parser.sections():
+        if len(existing_files) == 1:
+            return [parser]
+        parsers = []
+        parsers.append(parser)  # first parser is the one with all the settings.
+        for config_file in existing_files:
+            single_parser = ConfigParser()
+            single_parser.read(config_file, encoding='cp1250')
```
facundobatista: Aren't we parsing the files *twice*? Why not directly parse the config_files one by one and return them?

fades/file_options.py
```diff
@@ -0,0 +1,94 @@
...
+    """Merge options in MERGEABLE_CONFIGS in a uniq ConfigParser."""
+    if len(parsers) == 1:
+        return parsers[0]
+    merged_parser = parsers[0]
+    for parser in parsers[1:]:
+        # list of parsers is ordered by priority, last one is highest.
+        items = parser.items('fades')
+        for key, value in items:
+            if key in MERGEABLE_CONFIGS:
+                new_value = "{},{}".format(merged_parser.get('fades', key), value)
```
facundobatista: if the key is not there yet, the .get() crashes...

fades/file_options.py
```diff
@@ -0,0 +1,94 @@
...
+    parser = merge_parsers(parsers)
+    config_items = parser['fades']
+    for config_key, config_value in config_items.items():
+        if config_value == ['true', 'false']:
+            config_value = config_items.getboolean(config_key)
+        if config_key in MERGEABLE_CONFIGS:
+            current_value = getattr(args, config_key, [])
+            if current_value is None:
+                current_value = []
+            config_value = config_value.split(',')
```
facundobatista: isn't this a problem if I want a dependency like    "requests>=3,<4"    ?

fades/file_options.py
```diff
@@ -0,0 +1,94 @@
...
+
+    Config files will be parsed with priority equal to his order in CONFIG_FILES.
+    """
+    parsers = get_parsers()
+    if parsers is None:
+        return args
+    parser = merge_parsers(parsers)
+    config_items = parser['fades']
+    for config_key, config_value in config_items.items():
+        if config_value == ['true', 'false']:
```
facundobatista: equal to the list?=

README.rst
gilgamezh: Are you sure? I was thinking in config files like .vimrc .bashrc .gitconfig .tmux.conf, etc.

Also I should use `os.path.expanduser('~')`

README.rst
gilgamezh: fixed

fades/file_options.py
gilgamezh: fixed

README.rst
gilgamezh: fixed

fades/file_options.py
gilgamezh: idem https://github.com/PyAr/fades/pull/169#discussion_r51508838

https://docs.python.org/3.5/library/os.path.html#os.path.expanduser

> On Unix and Windows, return the argument with an initial component of ~ or ~user replaced by that user‘s home directory.

fades/file_options.py
gilgamezh: I'm doing it so that ConfigParser does the work of overwriting the "no mergeable" settings.

fades/file_options.py
gilgamezh: it's true. but I know that is there because I got it from the same place/file.

fades/file_options.py
gilgamezh: +1

fades/file_options.py
gilgamezh: my bad. I have to change it.

fades/file_options.py
gilgamezh: :hankey: 

I'm checking why the tests weren't failing.

fades/file_options.py
gilgamezh: I don't know what to do with the /etc files...   I will find if I find a solution

facundobatista: Maybe it's just ok... in Linux you may find /etc/fades.ini or not, in Windows you surely won't find it, no harm...

fades/file_options.py
gilgamezh: related to #149

PR #187, opened 2016-03-25
Change Version to Tuple

### LINKED ISSUE #175: Include also a VERSION attribute with the separated numbers
So, if fades.__version__ is ```"4.9"``` (which is ok), VERSION should be ```(4, 9)```, which is easier to compare programmatically.

Probably we should convert the internal _version module to be the tuple, and convert it to text where it's currently used....

dlitvakb:
Fixes #175

### DIFF
--- a/fades/__init__.py
+++ b/fades/__init__.py
@@ -17,6 +17,6 @@
 
 """Main package."""
 
-from ._version import __version__  # NOQA; provides module level version attr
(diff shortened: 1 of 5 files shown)


facundobatista:
Talked with David about *really* including the VERSION attribute.

facundobatista:
Awesome, thanks!!!

PR #188, opened 2016-03-25
Remove VirtualEnv on dependency Failure

### LINKED ISSUE #148: Cleanup just created venv if something fails during installation and everything explodes
Typical case:

    fades -v -d request==9.5

it will crash (no such version for that dependency), but a venv dir will be left there, totally not useful.

dlitvakb:
Fixes #148


facundobatista:
Did a first review, talked with David to include in the test an assertion that destroy_venv was actually called.

dlitvakb:
@facundobatista done!

facundobatista:
Awesome, please re merge from master so we can land this, thanks!

Issue #170, opened 2016-02-02
Have a way to clean old venvs


facundobatista:
We need to add an option:

    --clean-unused-venvs=X

It will remove cached virtualenvs (from internal structure and from disk) that were not used in the last X days.

To avoid writing the whole main metadata file everytime a venv is used, we'll have a separate "usage" metadata file.

Everytime fades is used, it will add a new line to this file: venv id and timestamp, very simple, very fast. There is an special case here, though: if the file is not found (or un-openable), it will add one line for each existent venv with current time (this will support first execution of the code, or if file is removed somehow, etc).

When fades is executed with --clean-unused-venvs, it will read the whole file, remove old venvs, and rewrite the file leaving one entry for each venv (with last timestamp in each case). If something is wrong with the file (not there, corrupt, etc), it will just avoid removing anything and leave the file in a reset status (one entry per venv with current timestamp).

We should recommed putting this option in a crontab. Maybe actually doing that for the user?

PR #190, opened 2016-03-27
Clean old venvs

arielrossanigo:
Hi, this PR is related to issue #170. 

With this pull request i've added this behaviors:
* When fades is invoked, always test the existence of usage_stats file. If it doesn't exists, it's created and filled with all the current venvs in cache.

* When venv it's used one line is recorded in usage file. This line contains the venv's uuid and datetime of usage.

* When fades is called with the option *--clean-unused-venvs=X*; after the first check, the usage file is compacted and all venvs that haven't been used for more than X days are removed. That's mean, remove from usage stat file, from index and the fisical env.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -164,10 +164,6 @@ If you want to use IPython shell you need to call *fades* with ``-i`` or
 ``--ipython`` option. This option will add IPython as a dependency to *fades*
 and it will launch this shell instead of the python one.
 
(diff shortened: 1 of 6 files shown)


facundobatista:
Did a first review, talked with Ariel to fix some details.

facundobatista:
Awesome, thanks!!!

Issue #193, opened 2016-03-27
Don't create a FadesEnvBuilder from envbuilder.destroy_env


facundobatista:
It should do ```FadesEnvBuilder.destroy_env(env_path)```, as a staticmethod or something, and not going through its ```__init__.py```

PR #196, opened 2016-03-29
destroy_env now is a static method instead of a instance one

arielrossanigo:
Hi, this is a fix to the issue #193.


facundobatista:
Thanks!!!

PR #197, opened 2016-03-29
if no cache in params, then only remove folder. Fix test

arielrossanigo:
When a dependency fail to be installed, the process call to destroy_venv with no cache parameter and fails.

This problem was not detected by tests because the assert was in the same scope where the exception it's captured.


facundobatista:
Thanks! Interfaces looks fluffy, but opened #200 to deal with this

PR #205, opened 2016-05-16
Improve the env building interfaces. Fixes #200.

### LINKED ISSUE #200: Refactor venv "cache" and "disk manager" interaction
Today, we have a "venvscache" and a pair of "create venv" and "destroy venv" in other module.

It feels there should be only one interface about this.

facundobatista:
Studied the relation between venvscache, venvbuilder, and who used them.

VEnvsCache is the cache, only, and shouldn't have action over the rest of the system.

I initially thought that we could move more stuff into FadesEnvBuilder, but not really, the best interface was to have creation and removing at module's level. So why to have FadesEnvBuilder as a class??? well, just to reuse virtualenv's EnvBuilder, so I just changed it's name to make it private in the module and improved its docstring.

Also, moved out the path removal from FadesEnvBuilder to the module's destroy_venv, which is the preferred interface.

### DIFF
--- a/fades/envbuilder.py
+++ b/fades/envbuilder.py
@@ -38,8 +38,13 @@ from fades.multiplatform import filelock
 logger = logging.getLogger(__name__)
 
 
-class FadesEnvBuilder(EnvBuilder):
-    """Create always a virtualenv."""
+class _FadesEnvBuilder(EnvBuilder):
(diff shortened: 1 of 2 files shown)


gilgamezh:
:+1:

PR #207, opened 2016-05-19
Create base directories if nonexistant. Several tests added.

### LINKED ISSUE #204: Need to assert that the requested directories exist
The base and config directories should be checked to exist; if they don't, create them.

facundobatista:
Create base directories if nonexistant. Several tests added. Fixes #204.


gilgamezh:
+1

PR #210, opened 2016-06-16
Install from repositories; fixes #7

### LINKED ISSUE #7: Support branching/installing directly from repositories
This is:
 - github
 - launchpad
 - bitbucket

We need a way to mark (in the import) not only the repository, but also the protocol used (e.g., bitbucket can work with hg or git)

facundobatista:


gilgamezh:
thanks! 

I can install from git private repositories now :+1:

PR #211, opened 2016-06-16
Simple useful scripts for when developing

facundobatista:
I stopped handcrafting the commands and just saved the files, if you too think they're useful...


facundobatista:
Superseeded by other cleaner one

PR #214, opened 2016-07-13
Fix really important error about --pip-options argument

humitos:
I always wanted to be part of this amazing project!


gilgamezh:
gracias!

PR #220, opened 2016-07-21
Add utf-8 encoding declaration to all files

Mariocj89:
Docstring of the files contain a name with non ascii character requires to add the encoding declaration

Otherwise, trying to run fades gives:
```
$ fades -d django
  File "/usr/local/bin/fades", line 4
SyntaxError: Non-ASCII character '\xc3' in file /usr/local/bin/fades on line 4, but no encoding declared; see http://python.org/dev/peps/pep-0263/ for details
```

### DIFF
--- a/bin/fades
+++ b/bin/fades
@@ -1,4 +1,5 @@
 #!/usr/bin/env python3
+# -*- coding: utf-8 -*-
 
 #
 # Copyright 2014 Facundo Batista, Nicolás Demarchi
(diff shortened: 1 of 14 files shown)


facundobatista:
'fades' is a Python 3 only project, which doesn't need that declaration in each file.

We don't support fades being executed by Python 2 (note that we support fades creating virtualenvs with python2 for other dependencies, if you need that, but fades itself should always be executed by Python 3)

So, your problem arises probably because you installed fades using Python 2?

PR #225, opened 2016-07-21
Useful scripts, again (?)

facundobatista:
It looks that originally they weren't included...


facundobatista:
Just merged the PR, don't know how we messed up the other branch that scripts didn't end there...

PR #227, opened 2016-07-23
Issue a WARNING if fades is executed from a virtualenv. Fixes #226.

### LINKED ISSUE #226: fades should issue a WARNING when it's executed from inside a virtualenv
As we don't support it, at least let the user know that something weird is going on

facundobatista:


facundobatista:
This branch was done with a lot of help from @blueyed

blueyed:
Awesome, thanks! :+1:

fades/main.py
```diff
@@ -63,6 +63,32 @@ def _merge_deps(*deps):
     return final
 
 
+def detect_inside_virtualenv():
+    """Tell if fades is inside a virtualenv.
+
+    This is a slightly adapted version from ipython's own check.
+    """
+    p_venv = os.environ.get('VIRTUAL_ENV')
```
gilgamezh: I think that the best way to check if is running inside a venv is with sys.prefix. With the environment variable we are not covering all the cases. 

More: https://www.python.org/dev/peps/pep-0405/#specification
Example: https://github.com/pypa/pip/blob/281eb61b09d87765d7c2b92f6982b3fe76ccb0af/pip/locations.py#L39

facundobatista:
We spent quite some time analysing different situations, going from the code of IPython, and found that the current code complies with everything.

Do you mind specifying which cases are not covered?

All this said, I think that the solution that pip uses is more inline with the venv PEP you linked. I just tried several cases too (possibly not all the ones we tried some weeks ago) and it works ok.

So Iĺl change this using your suggestion.

gilgamezh:
I think that the case is when de venv is not activated. (Because the VIRTUAL_ENV variable is setted  by the `activate` script)


```
11:51:09 |  /tmp  cat sarasa/bin/activate | grep VIRTUAL_ENV=
VIRTUAL_ENV="/tmp/sarasa"
11:51:10 |  /tmp  sarasa/bin/activate
zsh: permission denied: sarasa/bin/activate
11:51:31 |  ✘  /tmp  source sarasa/bin/activate
(sarasa) 11:52:07 |  /tmp  python test_venv.py 
/tmp/sarasa
(sarasa) 11:52:12 |  /tmp  deactivate 
11:52:14 |  /tmp  sarasa/bin/python test_venv.py 
  File "test_venv.py", line 4, in <module>
    print(os.environ['VIRTUAL_ENV'])
  File "/tmp/sarasa/lib/python3.5/os.py", line 725, in __getitem__
    raise KeyError(key) from None
KeyError: 'VIRTUAL_ENV'
11:53:17 |  ✘  /tmp  cat test_venv.py 

import os

print(os.environ['VIRTUAL_ENV'])
11:53:23 |  /tmp 
```

gilgamezh:
gracias! +1

blueyed:
gracias from me, too! :)

Issue #232, opened 2016-09-13
show the fades version in this string


gilgamezh:
`*** fades ***  2016-09-13 06:15:30,504  INFO     Need to install a dependency with pip, but no builtin, doing it manually (just wait a little, all should go well)`

PR #234, opened 2016-09-13
say hi on first call of pipmanager.install

luciotorre:
make sure we show the version if we are going to log something.

feel free to make suggestions for more reasonable strings.


gilgamezh:
resolve #232

facundobatista:
This is a very good idea!

What about ``"Hi! This is fades {}, automatically managing your dependencies".format(__version__)``?

facundobatista:
Awesome, thanks to both!

PR #243, opened 2017-03-25
Cleanup if pip download fails

facundofc:
In the case pip isn't installed on the current virtualenv, the PipManager downloads a `get-pip.py` script that bootstraps it. If you cancel the download or something goes wrong, the file gets created but might be incomplete. If you retry, you get an error because it asumes it already has the `get-pip.py`.

This pull request makes the file downloading more atomic by first downloading to a temp file and only after complete copying to the definitive file.

### DIFF
--- a/fades/pipmanager.py
+++ b/fades/pipmanager.py
@@ -59,7 +59,7 @@ class PipManager():
         if not self.pip_installed:
             logger.info("Need to install a dependency with pip, but no builtin, "
                         "doing it manually (just wait a little, all should go well)")
-            self.brute_force_install_pip()
+            self._brute_force_install_pip()
 
         str_dep = str(dependency)
         args = [self.pip_exe, "install", str_dep]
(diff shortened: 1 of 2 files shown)


fades/pipmanager.py
```diff
@@ -87,16 +90,22 @@ def get_version(self, dependency):
                          'Run with -v or check the logs for details')
             return ''
 
-    def _brute_force_install_pip(self):
+    def download_pip_installer(self):
+        u = request.urlopen(PIP_INSTALLER)
+        with contextlib.closing(u), \
```
facundobatista: Column limit in the project is 99c, no need for the \ here

fades/pipmanager.py
```diff
@@ -87,16 +90,22 @@ def get_version(self, dependency):
...
+                tempfile.NamedTemporaryFile('wb') as f:
+            shutil.copyfileobj(u, f)
+            f.flush()
+            shutil.copy(f.name, self.pip_installer_fname)
+
```
facundobatista: "copy" is not an atomic operation, you need to use os.rename() here.

For the rename to really work, you need the file in the same partition, so you have two choices here: or constrain the NamedTemporaryFile call for that, or just use something like self.pip_installer_fname + ".temp"

fades/pipmanager.py
```diff
@@ -87,16 +90,22 @@ def get_version(self, dependency):
                          'Run with -v or check the logs for details')
             return ''
 
-    def _brute_force_install_pip(self):
+    def download_pip_installer(self):
```
facundobatista: shouldn't this method be private?

tests/test_pipmanager.py
```diff
@@ -110,3 +112,38 @@ def test_say_hi_on_first_install(self):
...
+    def test_download_pip_installer(self):
+        mgr = PipManager('/usr/bin', pip_installed=False)
+        with patch('fades.pipmanager.request.urlopen') as urlopen, \
+                patch('shutil.copy') as copy, \
+                patch('shutil.copyfileobj') as copyobj:
+            mgr.download_pip_installer()
+
+        urlopen.assert_called_once_with(pipmanager.PIP_INSTALLER)
+        self.assertEqual(copyobj.call_count, 1)
+        copy.assert_has_calls([mock.call(mock.ANY, mgr.pip_installer_fname)])
```
facundobatista: Generally, for the whole tests (let's tallk IRL a little about this)...

I don't like patching those low level stdlib calls... for example, patching os.path.exists, it's a kind of mess if the tests infrastructure itself uses (now or in the future) that call while you have it patched.

For these kind of tests, what I would do is to patch the filename that the code will use to make it use some file in a temporary dict. Then, you could try the code if the file is there, if the file is not there, etc. The only thing I would patch is request.urlopen (we can be sure that the unittest infrastructure will not hit the network!).

The only test that may be tricky to do is to simulate the intermediate failure when downloading the code. However, I think that if you patch request.urlopen and make it raise an exception, it would fit for this (the file will not have some stuff in it, but it will be created (even if empty), and it will simulate the problem)

PR #249, opened 2017-04-09
add the virtualenv /bin path to the child PATH. fixes #248

### LINKED ISSUE #248: add virtualenv /bin path to the child PATH
When you are running a child program with fades and its program is doing some syscall the virtualenv PATH isn't in the process PATH. so It will fail to find a executable installed as a dependency.

IE:  nikola is calling ghp-import here: https://github.com/getnikola/nikola/blob/master/nikola/plugins/command/github_deploy.py#L50

I'm calling nikola using this alias `nikola='fades -d nikola -d ghp-import2 -d micawber -x nikola'` 

And I'm getting: 

```
[2017-04-09T19:33:23Z] ERROR: Nikola: In order to deploy the site to GitHub Pages, you must install the "ghp-import2" Python package.```

### PR DESCRIPTION
**Pull request by gilgamezh**

### CODE REVIEW

File: fades/main.py [hunk 1]
```diff
@@ -260,6 +260,9 @@ def go(argv):
     # run forest run!!
     python_exe = 'ipython' if args.ipython else 'python'
     python_exe = os.path.join(venv_data['env_bin_path'], python_exe)
+    # add the virtualenv /bin path to the child PATH.
```
facundobatista: please leave an empty line above the comment so we have a "block of code" (it's the style of the rest of the text)

fades/main.py
```diff
@@ -260,6 +260,9 @@ def go(argv):
...
+    os.environ['PATH'] = "{}:{}".format(venv_data['env_bin_path'], os.environ.get('PATH'))
+    logger.debug("child PATH: %s", os.environ)
```
facundobatista: The logger in this code is called "l" (feel free to use this branch to enhance this name), not logger (note that logger is the module we imported).

Also, please start the logging line in uppercase.

Finally, don't log the WHOLE os.environ, just the PATH value

fades/main.py
```diff
@@ -260,6 +260,9 @@ def go(argv):
     # run forest run!!
     python_exe = 'ipython' if args.ipython else 'python'
     python_exe = os.path.join(venv_data['env_bin_path'], python_exe)
+    # add the virtualenv /bin path to the child PATH.
+    os.environ['PATH'] = "{}:{}".format(venv_data['env_bin_path'], os.environ.get('PATH'))
```
facundobatista: If PATH is not in the environment, the `.get()` will return None, which will suck in the built path (it wil be `"....:None"`).

You could do `os.environ.get('PATH', "")`, but in the case of missing PATH you will end with a path that ends in the ":", would that be ok?

Otherwise you'd need to do something like: 
```
    environ_path = venv_data['env_bin_path']
    if 'PATH' in os.environ:
        environ_path += ':' + os.environ['PATH']
    os.environ['PATH'] = environ_path
```

fades/main.py
gilgamezh: done!

fades/main.py
gilgamezh: 👍

fades/main.py
gilgamezh: mmm we already have the venv path logged. I'll remove this line

gilgamezh: we are using `logger` everywhere, so I think is better to change the imported module nanme

PR #262, opened 2017-06-23
Two small details in project's organization.

facundobatista:
- Moved "debian" directory inside pkg, and updated the HOWTO

- Changed the condition in the fades script to fix the path not only if dir is "fades" but also if dir starts with "fades", which will include when you run fades from an opened tarball (that is normally called fades-version)


gilgamezh:
gracias!

PR #265, opened 2017-06-24
Log platform info using platform.

juancarlospaco:
- Change `sys.platform` to `platform.platform()` inside a `print()` for `--version`.
- 3 lines change, but logs better info about the system on the log, minor, tests pass Ok.

> Returns a single string identifying the underlying platform with as much useful information as possible.
> The output is intended to be human readable.

**Before:**
```python
System: linux
```

**After:**
```python
System: Linux-4.11.5-1-ARCH-x86_64-with-Archlinux
```

### DIFF
--- a/fades/main.py
+++ b/fades/main.py
@@ -19,6 +19,7 @@
 
 import argparse
 import os
+import platform
 import signal
 import sys
 import subprocess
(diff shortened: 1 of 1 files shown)


gilgamezh:
muchas gracias!

facundobatista:
I like it! Before landing this, would you please add yourself to the AUTHORS file in this same branch?

Thanks!

facundobatista:
Ah, I see it was already merged... mind to create a new branch, then, adding yourself to the AUTHORS file? Thanks!

PR #269, opened 2017-06-28
Ordered AUTHORS and included Juan Carlos Paco.

facundobatista:
I wonder if we should include something like the following in a test case:

```
$ fades -d pyuca
Python 3.5.2+ (default, Sep 22 2016, 12:18:14) 
[GCC 6.2.0 20160927] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from pyuca import Collator
>>> authors = open('AUTHORS', 'rt', encoding='utf8').readlines()
>>> authors == sorted(authors, key=Collator().sort_key)
True

```


gilgamezh:
@facundobatista +1 to do that. should be really easy to add in the test script

PR #273, opened 2017-07-16
add flake8 config file (it works with lot of text editors plugins)

gilgamezh:


facundobatista:
Merged, but what would you think to get this info from the infrastructure test case and not having defined twice?

For clarity, I'm suggesting that in the test case we should get the config for flake8 from this file you added

PR #274, opened 2017-07-16
check with PyPI before start to installing stuff. resolve #178

### LINKED ISSUE #178: It should check with PyPI before start to installing stuff
So, fades decides that the venv required doesn't exist, it starts to install a lot of stuff. 

Before starting to download and install a lot of stuff, it should check with the backend that _all_ the packages are there, so it won't fail after downloading a lot of stuff and installing it.

gilgamezh:


fades/helpers.py
```diff
@@ -208,3 +210,37 @@ def check_pypi_updates(dependencies):
...
+    else:
+        if response.status == HTTP_STATUS_OK:
+            logger.debug("%s exists in Pypi.", pkg_name)
+            return True
+
+
+def check_pypi_exists(dependencies):
+    """Check if the indicated dependencies actually exists in pypi."""
+    for dependency in dependencies.get('pypi', []):
+        logger.debug("Checking if %s exists in PyPi", dependency)
```
facundobatista: I'd use %r, just in case, because we may be hitting the issue of "hey, it fails to install dependency foobar why?" and the problem being that somehow the dependency verified is "foobar ", not "foobar"

fades/helpers.py
```diff
@@ -208,3 +210,37 @@ def check_pypi_updates(dependencies):
 
     dependencies["pypi"] = dependencies_up_to_date
     return dependencies
+
+
+def _pypi_head_package(pkg_name):
```
facundobatista: In this case we're not checking if the version exists, right? I mean, we're validation that dependency "foobar" exists, but what about if user requested foobar==2.3 and 2.3 doesn't exist?

fades/helpers.py
```diff
@@ -208,3 +210,37 @@ def check_pypi_updates(dependencies):
...
+    url = BASE_PYPI_URL.format(name=pkg_name)
+    req = request.Request(url, method='HEAD')
+    try:
+        response = request.urlopen(req)
+    except HTTPError as http_error:
+        if http_error.code == HTTP_STATUS_NOT_FOUND:
+            return False
+        else:
+            raise http_error
+    else:
```
facundobatista: considering that in the "except" never goes through (or it returns or it raises an exception), I'll kill this else and deindent the code inside it.

fades/helpers.py
```diff
@@ -208,3 +210,37 @@ def check_pypi_updates(dependencies):
...
+        try:
+            exists = _pypi_head_package(dependency.project_name)
+        except Exception as error:
+            logger.error("Error checking %s in pypi: %r", dependency, error)
+            sys.exit(1)
+        else:
+            if not exists:
+                logger.error("%s doesn't exists in PyPi.", dependency)
+                return False
+    logger.debug("All indicated dependencies exists in pypi=%s", dependencies)
```
facundobatista: you already logged a line for each dependency; is this one necessary? what value is it providing?

fades/main.py
```diff
@@ -116,6 +116,8 @@ def go(argv):
                               "used multiple times)"))
     parser.add_argument('--check-updates', action='store_true',
                         help=("check for packages updates"))
+    parser.add_argument('--no-precheck-availability', action='store_true',
+                        help=("Not check if the packages exists in pypi."))
```
facundobatista: s/Not/Don't/
s/pypi/PyPI/

fades/helpers.py
```diff
@@ -208,3 +210,37 @@ def check_pypi_updates(dependencies):
...
+    for dependency in dependencies.get('pypi', []):
+        logger.debug("Checking if %s exists in PyPi", dependency)
+        try:
+            exists = _pypi_head_package(dependency.project_name)
+        except Exception as error:
+            logger.error("Error checking %s in pypi: %r", dependency, error)
+            sys.exit(1)
+        else:
+            if not exists:
+                logger.error("%s doesn't exists in PyPi.", dependency)
```
facundobatista: I'd log this in INFO, because is not our error or from the system. I thought about the user calling fades with --quiet and not seeing this, but note that we're already logging in error below, where we say "some dependency is not there, exiting"

tests/test_helpers.py
```diff
@@ -29,6 +29,7 @@
 
 from xdg import BaseDirectory
 
+from fades import HTTP_STATUS_NOT_FOUND, HTTP_STATUS_OK
 from fades import helpers
 from fades import parsing
```
facundobatista: I wonder why don't we have all this imports in the same line...

fades/helpers.py
gilgamezh: that's totally true. Let's discuss this on the IRC

fades/helpers.py
gilgamezh: +1

fades/helpers.py
gilgamezh: +1

fades/main.py
gilgamezh: +1

tests/test_helpers.py
gilgamezh: my bad :-) I will change it

gilgamezh:
@facundobatista All requested changes already done. (sorry for the delay)

fades/helpers.py
gilgamezh: Now it's checking with the version.

fades/helpers.py
gilgamezh: done

fades/helpers.py
gilgamezh: done

fades/helpers.py
gilgamezh: done

fades/helpers.py
gilgamezh: removed

fades/main.py
gilgamezh: done

tests/test_helpers.py
gilgamezh: done

fades/helpers.py
```diff
@@ -208,3 +211,40 @@ def check_pypi_updates(dependencies):
...
+
+
+def _pypi_head_package(dependency):
+    """Hit pypi with a http HEAD to check if pkg_name exists."""
+    if dependency.specs:
+        _, version = dependency.specs[0]
+        url = BASE_PYPI_URL_WITH_VERSION.format(name=dependency.project_name, version=version)
+    else:
+        url = BASE_PYPI_URL.format(name=dependency.project_name)
+    logger.debug("Doing HEAD requests agains %s", url)
```
facundobatista: s/agains/against/

fades/helpers.py
```diff
@@ -208,3 +211,40 @@ def check_pypi_updates(dependencies):
...
+    req = request.Request(url, method='HEAD')
+    try:
+        response = request.urlopen(req)
+    except HTTPError as http_error:
+        if http_error.code == HTTP_STATUS_NOT_FOUND:
+            return False
+        else:
+            raise http_error
```
facundobatista: It's better to just "raise" here, because this way it will let propagate the original exception, and not raise another one.

fades/helpers.py
```diff
@@ -208,3 +211,40 @@ def check_pypi_updates(dependencies):
...
+    if response.status == HTTP_STATUS_OK:
+        logger.debug("%r exists in Pypi.", dependency)
```
facundobatista: s/Pypi/PyPI/

fades/helpers.py
```diff
@@ -208,3 +211,40 @@ def check_pypi_updates(dependencies):
...
+        return True
+
```
facundobatista: I wonder... what if there was not an http_error, but status is not OK, what would or should happen?

IOW, there is a missing "else" here.

Maybe we could consider that if we're without an exception, it's OK, and just return True? (without checking equality against HTTP_STATAUS_OK)

fades/helpers.py
```diff
@@ -208,3 +211,40 @@ def check_pypi_updates(dependencies):
...
+
+def check_pypi_exists(dependencies):
+    """Check if the indicated dependencies actually exists in pypi."""
```
facundobatista: s/pypi/PyPI/

fades/helpers.py
```diff
@@ -208,3 +211,40 @@ def check_pypi_updates(dependencies):
...
+    for dependency in dependencies.get('pypi', []):
+        logger.debug("Checking if %r exists in PyPi", dependency)
```
facundobatista: s/PyPi/PyPI/

fades/helpers.py
```diff
@@ -208,3 +211,40 @@ def check_pypi_updates(dependencies):
...
+        try:
+            exists = _pypi_head_package(dependency)
+        except Exception as error:
+            logger.error("Error checking %s in pypi: %r", dependency, error)
```
facundobatista: s/pypi/PyPI/

fades/helpers.py
```diff
@@ -208,3 +211,40 @@ def check_pypi_updates(dependencies):
...
+            sys.exit(1)
+        else:
+            if not exists:
+                logger.info("%s doesn't exists in PyPi.", dependency)
```
facundobatista: s/PyPi/PyPI/

fades/main.py
```diff
@@ -253,6 +255,12 @@ def go(argv):
         create_venv = True
 
     if create_venv:
+        # Check if the requested packages exists in pypi.
+        if not args.no_precheck_availability:
+            if not helpers.check_pypi_exists(indicated_deps):
```
facundobatista: When we do this checking, we are really hitting the network. Probably we should log in INFO before doing that? something like "Checking the availabilty of dependnecies in PyPI"?

fades/main.py
```diff
@@ -253,6 +255,12 @@ def go(argv):
...
+                logger.info("A indicated dependency doesn't exists. Exiting")
```
facundobatista: Some details here:

- s/A indicated/An indicated/

- this situation is serious enough to make fades quit... maybe we should log in WARNING?

- wouldn't it be great if we could log here (or in the called function) *which dependency* is the one generating this situation?

gilgamezh: - typo fixed 
- I think if the user is asking for a dependency that doesn't exists we should fail ASAP. (it's maybe a typo) 
- we are currently showing it 

```
 ~/code/fades   precheck-availability ●  bin/fades -d sarasa                                                                                                 123476ms 
*** fades ***  2018-03-17 18:17:30,636  INFO     sarasa doesn't exists in PyPI.
*** fades ***  2018-03-17 18:17:30,637  INFO     An indicated dependency doesn't exists. Exiting
```

gilgamezh: FTR: I will change the loglevel to ERROR

fades/helpers.py
gilgamezh: done

fades/helpers.py
gilgamezh: +1 done

fades/helpers.py
gilgamezh: done

fades/helpers.py
gilgamezh: I added a "else" and a log in warning.

fades/helpers.py
gilgamezh: done

fades/helpers.py
gilgamezh: done

fades/helpers.py
gilgamezh: done

fades/helpers.py
gilgamezh: done

fades/main.py
gilgamezh: +1

fades/main.py
```diff
@@ -253,6 +255,14 @@ def go(argv):
         create_venv = True
 
     if create_venv:
+        # Check if the requested packages exists in pypi.
+        if not args.no_precheck_availability:
+            logger.info("Checking the availabilty of dependencies in PyPI. "
+                        "You can use '--no_precheck_availability' to avoid it.")
```
facundobatista: it's '--no-precheck-availability'

PR #275, opened 2017-07-16
Tell Popen to already return unicode.

facundobatista:


facundobatista:
Fixed the original decoding issue for windows case (in a multiplatform way) and restructured the whole temp files usage in the tests, because windows complained about files being locked when wanted to be used.

I think this branch should be merged, but note all problems are not solved :(

As you see in the [last appveyor run](https://ci.appveyor.com/project/facundobatista/fades/build/1.0.34), it fails for Python 3.4 64b. Very strange, because the rest works.

But as I've just said, let's merge this and then decide what to do with that issue...

gilgamezh:
gracias!

PR #278, opened 2017-09-17
use newer versions of dev dependencies

gilgamezh:


facundobatista:
Did you see that all Travis failed because a flake8 api change?

gilgamezh:
I'll create a new (updated) PR

PR #285, opened 2018-03-18
don't duplicate dependencies by repo

### LINKED ISSUE #279: Fades installs the same dependency multiple times?
In my application, I have:

```python
from prompt_toolkit import prompt
from prompt_toolkit.history import FileHistory
from prompt_toolkit.auto_suggest import AutoSuggestFromHistory
from prompt_toolkit.contrib.completers import WordCompleter
```

If I put `# fades` at the end of each of those lines, fades will tell me:

```
*** fades ***  2017-10-10 00:32:35,603  INFO     Hi! This is fades 6.0, automatically managing your dependencies                                                             
*** fades ***  2017-10-10 00:32:35,610  INFO     Installing dependency: 'prompt-toolkit'                                                                                     
*** fades ***  2017-10-10 00:32:50,852  INFO     Installing dependency: 'prompt-toolkit'                                                                                     
*** fades ***  2017-10-10 00:32:58,400  INFO     Installing dependency: 'prompt-toolkit'                                                                                     
*** fades ***  2017-10-10 00:33:06,013  INFO     Installing dependency: 'prompt-toolkit'
```

Is it really installing the same prompt-toolkit module 4 times?  Shouldn't one time be enough?

Also, it's not clear from the documentation if `# fades` is necessary on each such module import line, or only necessary on the first one (if there are multiple imports from the same module, as in my case).

gilgamezh:
resolves #279 

:warning: Whit this solution we are only avoiding duplicated by repo.

### DIFF
--- a/fades/main.py
+++ b/fades/main.py
@@ -60,7 +60,7 @@ def _merge_deps(*deps):
     final = {}
     for dep in deps:
         for repo, info in dep.items():
-            final.setdefault(repo, []).extend(set(info))
+            final.setdefault(repo, set()).update(info)
     return final
 
 
--- a/fades/main.py
+++ b/fades/main.py
@@ -60,7 +60,7 @@ def _merge_deps(*deps):
     final = {}
     for dep in deps:
         for repo, info in dep.items():
-            final.setdefault(repo, []).extend(info)
+            final.setdefault(repo, []).extend(set(info))
     return final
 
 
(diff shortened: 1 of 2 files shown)


fades/main.py
```diff
@@ -60,7 +60,7 @@ def _merge_deps(*deps):
     final = {}
     for dep in deps:
         for repo, info in dep.items():
-            final.setdefault(repo, []).extend(info)
+            final.setdefault(repo, []).extend(set(info))
```
facundobatista: why not having, in final, per repo, directly a set? something like:
```
final.setdefault(repo, set()).update(info)
```

gilgamezh: I wasn't sure if the code is expecting to receive a list in any other place. 

I can check it later.

facundobatista:
Awesome, thanks

PR #286, opened 2018-03-19
delete temp folder used for tests

gilgamezh:
one tests wasn't deleting tmp files 

```
 /tmp  tree tmp*                                                                                                                                                    5ms 
tmp14aoscfc                                                                                                                                                              
└── usage_stats                                                                                                                                                          
tmp1nx0ux37                                                                                                                               
tmp1sjmitat                                                                                                                       
└── usage_stats                                                                                                                                                          
tmp43bvm3ux                                                                                                                                                              
└── usage_stats                                                                                                                                                          
tmp_87fw02m                                                                                                                                                              
└── usage_stats                                                                                                                                                          
tmpqk3u3p1g                                                                                                                                                              
tmpzz1nqf3n ```

### DIFF
--- a/tests/test_envbuilder.py
+++ b/tests/test_envbuilder.py
@@ -17,6 +17,7 @@
 """Tests for the venv builder module."""
 
 import os
+import shutil
 import tempfile
 import unittest
 from datetime import datetime, timedelta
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: tests/test_envbuilder.py [hunk 1]
```diff
@@ -278,6 +279,7 @@ def setUp(self):
         self.temp_folder = tempfile.mkdtemp()
         self.file_path = os.path.join(self.temp_folder, 'usage_stats')
         self.addCleanup(lambda: os.path.exists(self.tempfile) and os.remove(self.tempfile))
```
facundobatista: Why do we keep the cleanup of inside files? Isn't removing the whole dir enough?

gilgamezh: check the first line of `setUp` the test is creating two temp files.

PR #293, opened 2018-03-25
not crash if executable is not there. resolve #261

### LINKED ISSUE #261: fades should not crash if executable is not there
For example:

```
$ fades -d nose -x nosetest tests/test_extract.py
  File "/usr/bin/fades", line 33, in <module>
    main.go(sys.argv)
  File "/usr/lib/python3/dist-packages/fades/main.py", line 291, in go
    p = subprocess.Popen(cmd + args.child_options)
  File "/usr/lib/python3.5/subprocess.py", line 947, in __init__
    restore_signals, start_new_session)
  File "/usr/lib/python3.5/subprocess.py", line 1551, in _execute_child
    raise child_exception_type(errno_num, err_msg)
FileNotFoundError: [Errno 2] No such file or directory: '/home/facundo/.local/share/fades/4089d63f-62bd-4bac-a178-9cc04401248f/bin/nosetest'
```

It should log something like "ERROR - File not found in virtualenv bin dir". Maybe also indicate which files *are* there? Maybe show only similar ones? Probably also indicate the path of the venv.

gilgamezh:
It's tested in my local. I was trying to add some tests and I wasn't able to. 

tl;dr:  tests are running in a virtualenv. And fades doesn't support running in a venv. 

I think for the next release we should do some refactor in `go()` splitting it in smaller pieces to allow testing.
Then we can add tests for fades.main

### DIFF
--- a/fades/main.py
+++ b/fades/main.py
@@ -300,7 +300,11 @@ def go():
                 "Calling Python interpreter with arguments %s to execute the child program"
                 " %r with options %s", python_options, args.child_program, args.child_options)
 
-        p = subprocess.Popen(cmd + args.child_options)
+        try:
+            p = subprocess.Popen(cmd + args.child_options)
+        except FileNotFoundError:
+            logger.error("Command not found: %s", args.child_program)
+            sys.exit(1)
 
     def _signal_handler(signum, _):
         """Handle signals received by parent process, send them to child.


facundobatista:
Yes, we have #149 for that...

PR #313, opened 2018-04-21
Check for pip.exe beyond the normal pip binary.

facundobatista:


facundobatista:
You can even check that this is working ok in the AppVeyor automatic test; no more `pip isn't installed in the venv, setting pip_installed=False` there...

PR #318, opened 2018-04-30
Replace sys.exit calls with exceptions

andresdelfino:
Pass exceptions to the main module instead of sys.exiting.

### DIFF
--- a/fades/envbuilder.py
+++ b/fades/envbuilder.py
@@ -38,6 +38,14 @@ from fades.multiplatform import filelock
 logger = logging.getLogger(__name__)
 
 
+class FadesError(Exception):
+    """Provides a Fades exception."""
+
+    def __init__(self, reason):
+        """Init."""
+        self.reason = reason
+
+
 class _FadesEnvBuilder(EnvBuilder):
     """Create always a virtualenv.
 
@@ -84,13 +92,13 @@ class _FadesEnvBuilder(EnvBuilder):
         except FileNotFoundError as error:
             logger.error('Virtualenv is not installed. It is needed to create a virtualenv with '
                          'a different python version than fades (got {})'.format(error))
-            sys.exit(1)
+            raise FadesError('virtualenv not found')
         except helpers.ExecutionError as error:
             error.dump_to_log(logger)
-            sys.exit(2)
+            raise FadesError('virtualenv could not be run')
         except Exception as error:
             logger.exception("Error creating virtualenv:  %s", error)
-            sys.exit(3)
+            raise FadesError('General error while running virtualenv')
 
     def create_env(self, interpreter, is_current, options):
         """Create the virtualenv and return its info."""
@@ -149,7 +157,7 @@ def create_venv(requested_deps, interpreter, is_current, options, pip_options):
             except Exception:
                 logger.debug("Installation Step failed, removing virtualenv")
                 destroy_venv(env_path)
-                sys.exit(4)
+                raise FadesError('Dependency installation failed')
 
             if repo == REPO_VCS:
                 # no need to request the installed version, as we'll always compare
--- a/fades/envbuilder.py
+++ b/fades/envbuilder.py
@@ -30,20 +30,12 @@ from datetime import datetime
 from venv import EnvBuilder
 from uuid import uuid4
 
-from fades import REPO_PYPI, REPO_VCS
+from fades import FadesError, REPO_PYPI, REPO_VCS
 from fades import helpers
 from fades.pipmanager import PipManager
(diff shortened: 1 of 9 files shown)


fades/envbuilder.py
```diff
@@ -38,6 +38,14 @@
 logger = logging.getLogger(__name__)
 
 
+class FadesError(Exception):
+    """Provides a Fades exception."""
+
+    def __init__(self, reason):
```
facundobatista: This is not really needed, as Exceptions already receive and store a `message`, so just define FadesError, with the docstring, and that's all.

That said, it's better to define this exception in a more general place... put it in `fades/__init__.py` and import it from there...

fades/helpers.py
```diff
@@ -135,7 +135,7 @@ def _get_interpreter_info(interpreter=None):
             requested_interpreter_info = logged_exec(args)
         except Exception as error:
             logger.error("Error getting requested interpreter version: %s", error)
-            sys.exit(1)
+            raise
```
facundobatista: Use FadesError, as above (it's a good idea!!!)

fades/helpers.py
```diff
@@ -250,7 +250,7 @@ def check_pypi_exists(dependencies):
             exists = _pypi_head_package(dependency)
         except Exception as error:
             logger.error("Error checking %s in PyPI: %r", dependency, error)
-            sys.exit(1)
+            raise
```
facundobatista: Use FadesError...

fades/main.py
```diff
@@ -166,7 +166,7 @@ def go():
                                 getattr(sys, 'base_prefix', None)):
         logger.error(
             "fades is running from inside a virtualenv (%r), which is not supported", sys.prefix)
-        sys.exit(-1)
+        raise Exception
```
facundobatista: Use FadesError, as above

fades/main.py
```diff
@@ -181,13 +181,11 @@ def go():
     if args.clean_unused_venvs:
         try:
             max_days_to_keep = int(args.clean_unused_venvs)
-            usage_manager.clean_unused_venvs(max_days_to_keep)
-        except Exception:
-            rc = 1
-            logger.debug("CLEAN_UNUSED_VENVS must be an integer.")
+        except Exception.ValueError:
```
facundobatista: This is invalid, just use `ValueError`

fades/main.py
```diff
@@ -181,13 +181,11 @@ def go():
...
+            logger.debug("clean_unused_venvs must be an integer.")
```
facundobatista: Log in ERROR here, not DEBUG

fades/main.py
```diff
@@ -262,7 +260,7 @@ def go():
                         "You can use '--no-precheck-availability' to avoid it.")
             if not helpers.check_pypi_exists(indicated_deps):
                 logger.error("An indicated dependency doesn't exists. Exiting")
-                sys.exit(1)
+                raise Exception
```
facundobatista: Use FadesError

fades/main.py
```diff
@@ -304,7 +302,7 @@ def go():
             p = subprocess.Popen(cmd + args.child_options)
         except FileNotFoundError:
             logger.error("Command not found: %s", args.child_program)
-            sys.exit(1)
+            raise
```
facundobatista: Use FadesError

fades/main.py
```diff
@@ -181,13 +182,11 @@ def go():
     if args.clean_unused_venvs:
         try:
             max_days_to_keep = int(args.clean_unused_venvs)
-            usage_manager.clean_unused_venvs(max_days_to_keep)
-        except Exception:
-            rc = 1
-            logger.debug("CLEAN_UNUSED_VENVS must be an integer.")
+        except ValueError:
+            logger.error("clean_unused_venvs must be an integer.")
             raise
```
facundobatista: raise FadesError

fades/main.py
```diff
@@ -181,13 +182,11 @@ def go():
...
-        finally:
-            sys.exit(rc)
+
+        usage_manager.clean_unused_venvs(max_days_to_keep)
```
facundobatista: After this operation, fades must quit (this was handled in the `finally` you removed)

testdev.bat
```diff
@@ -0,0 +1,11 @@
+@echo off
+
+rem Copyright 2014-2016 Facundo Batista, Nicolás Demarchi
```
facundobatista: 2018 is enough here (this file is new)

facundobatista:
Awesome! Thank you very much!

Issue #320, opened 2018-04-30
It should accept -r multiple times


facundobatista:
This way the following could be done: `fades -r reqs.txt -r reqs_devel.txt`

andresdelfino:
Perhaps we should raise an exception when two requirement files mention the same package with different versions?

facundobatista:
We should grab all dependencies and mix them (not override anything).

Most of the difficult questions you ask are also present for the same file anyway (e.g.: what if I include requests==1 and request==2 in the same req.txt file?)

gilgamezh:
```
(test)  /tmp  pip install -r reqs1.txt -r reqs2.txt                                                                                                                 2ms 
Double requirement given: requests==2.18.3 (from -r reqs2.txt (line 1)) (already in requests==2.18.4 (from -r reqs1.txt (line 1)), name='requests')
You are using pip version 9.0.3, however version 10.0.1 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.
(test)  ✘  /tmp  pip install -r reqs2.txt -r reqs1.txt                                                                                                           407ms 
Double requirement given: requests==2.18.4 (from -r reqs1.txt (line 1)) (already in requests==2.18.3 (from -r reqs2.txt (line 1)), name='requests')
You are using pip version 9.0.3, however version 10.0.1 is available.
You should consider upgrading via the 'pip install --upgrade pip' command.
(test)  ✘  /tmp                                                          
```

andresdelfino:
I think it makes sense to follow pip's handling of these cases.

When there's a decision about this I'll be glad to implement it.

facundobatista:
Let's merge all dependencies and `pip` will decide if they make sense or something is colliding.

So the current PR is going in the good direction. Let's fix some stuff I requested there... Thanks!!!

andresdelfino:
Yes, I agree. If pip is catching whatever conflict that might happen, it's best not to duplicate that logic.

facundobatista:
Already in master! Thanks!!

PR #322, opened 2018-05-13
Multiple requirement files support

andresdelfino:
I'll update the PR with documentation patches after we come up with the proper solution.

### DIFF
--- a/fades/main.py
+++ b/fades/main.py
@@ -220,18 +220,18 @@ def go():
         docstring_deps = parsing.parse_docstring(args.child_program)
         logger.debug("Dependencies from docstrings: %s", docstring_deps)
 
-    unmerged_deps = [ipython_dep, indicated_deps, docstring_deps]
+    all_dependencies = [ipython_dep, indicated_deps, docstring_deps]
 
     for rf_path in args.requirement:
         rf_deps = parsing.parse_reqfile(rf_path)
-        logger.debug('Dependencies from requirements file "%s": %s', rf_path, rf_deps)
-        unmerged_deps.append(rf_deps)
+        logger.debug('Dependencies from requirements file %r: %s', rf_path, rf_deps)
+        all_dependencies.append(rf_deps)
 
     manual_deps = parsing.parse_manual(args.dependency)
     logger.debug("Dependencies from parameters: %s", manual_deps)
-    unmerged_deps.append(manual_deps)
+    all_dependencies.append(manual_deps)
 
-    indicated_deps = _merge_deps(unmerged_deps)
+    indicated_deps = _merge_deps(all_dependencies)
 
     # Check for packages updates
     if args.check_updates:
--- a/fades/main.py
+++ b/fades/main.py
@@ -219,15 +219,15 @@ def go():
         logger.debug("Dependencies from source file: %s", indicated_deps)
         docstring_deps = parsing.parse_docstring(args.child_program)
         logger.debug("Dependencies from docstrings: %s", docstring_deps)
-    reqfile_deps = {}
+    reqfile_deps = []
     for rfpath in args.requirement:
         rfdeps = parsing.parse_reqfile(rfpath)
         logger.debug('Dependencies from requirements file "%s": %s', rfpath, rfdeps)
-        reqfile_deps.update(rfdeps)
+        reqfile_deps.append(rfdeps)
     manual_deps = parsing.parse_manual(args.dependency)
     logger.debug("Dependencies from parameters: %s", manual_deps)
     indicated_deps = _merge_deps(ipython_dep, indicated_deps, docstring_deps,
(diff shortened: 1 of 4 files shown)


andresdelfino:
First step for closing #320.

fades/main.py
```diff
@@ -218,8 +219,11 @@ def go():
         logger.debug("Dependencies from source file: %s", indicated_deps)
         docstring_deps = parsing.parse_docstring(args.child_program)
         logger.debug("Dependencies from docstrings: %s", docstring_deps)
-    reqfile_deps = parsing.parse_reqfile(args.requirement)
-    logger.debug("Dependencies from requirements file: %s", reqfile_deps)
+    reqfile_deps = {}
+    for rfpath in args.requirement:
+        rfdeps = parsing.parse_reqfile(rfpath)
+        logger.debug('Dependencies from requirements file "%s": %s', rfpath, rfdeps)
+        reqfile_deps.update(rfdeps)
```
facundobatista: You can not do this, see that `parsing.parse_reqfile` returns a dict whose key is the "repo", so you would be overriding stuff from one req file into the other.

You need to get this info and use it below in the the `_merge_deps`  call.

What we *really* should do here is to extract the "dependencies grabbing" from inside here and put it somewhere else so it can be properly tested... but that is maybe work for another branch...

andresdelfino: I hope the update to the PR addresses your comment.

andresdelfino:
I changed _merge_deps to only accept an iterable. If this sounds ok, I'll update the test cases. Otherwise, if accepting multiple arguments is desired, I could test if a given argument is a list to iterate on the dictionaries it contains.

fades/main.py
```diff
@@ -58,8 +58,14 @@
 
 def _merge_deps(*deps):
     """Merge all the dependencies; latest dicts overwrite first ones."""
+    all_deps = []
     final = {}
     for dep in deps:
+        if isinstance(dep, dict):
```
facundobatista: This kind of "dual type" consumption is tricky... it makes the code a little more complex here, but makes usage of this function a LOT more complex, as it's difficult to remember how to use it.

So, let's keep it simple. This function should receive a  list where each list is the dependency dict. IOW, the only change I'd do in this function is to receive "deps" (without the star), and let's change below how we call this...

fades/main.py
```diff
@@ -100,8 +106,9 @@ def go():
     parser.add_argument('-d', '--dependency', action='append',
                         help="specify dependencies through command line (this option can be "
                              "used multiple times)")
-    parser.add_argument('-r', '--requirement',
-                        help="indicate from which file read the dependencies")
+    parser.add_argument('-r', '--requirement', action='append',
+                        help="indicate files to read dependencies from (this option can be "
```
facundobatista: Thanks for changing the help string here. But remember also to change the README and the man page to reflect this change.

fades/main.py
```diff
@@ -218,10 +225,16 @@ def go():
...
+    reqfile_deps = []
+    for rf_path in args.requirement:
+        rf_deps = parsing.parse_reqfile(rf_path)
+        logger.debug('Dependencies from requirements file "%s": %s', rf_path, rf_deps)
+        reqfile_deps.append(rf_deps)
+
     manual_deps = parsing.parse_manual(args.dependency)
     logger.debug("Dependencies from parameters: %s", manual_deps)
+
     indicated_deps = _merge_deps(ipython_dep, indicated_deps, docstring_deps,
```
facundobatista: Let's have a list around for all deps and that's it. This way we could call `_merge_deps(all_dependencies)`.

So start with an `all_dependencies` above and add the different dependencies from the different sources. One thing to pay attention is the order how these dependencies are stored in the list, it should respect:

1. ipython dep
1. deps from source file 
1. deps from file docstring
1. deps from requirement files
1. deps manually specified with `-d`

fades/main.py
```diff
@@ -218,10 +225,16 @@ def go():
         logger.debug("Dependencies from source file: %s", indicated_deps)
         docstring_deps = parsing.parse_docstring(args.child_program)
         logger.debug("Dependencies from docstrings: %s", docstring_deps)
-    reqfile_deps = parsing.parse_reqfile(args.requirement)
-    logger.debug("Dependencies from requirements file: %s", reqfile_deps)
+
+    reqfile_deps = []
+    for rf_path in args.requirement:
+        rf_deps = parsing.parse_reqfile(rf_path)
+        logger.debug('Dependencies from requirements file "%s": %s', rf_path, rf_deps)
```
facundobatista: Instead of `"%s"` you probably should just do `%r`.

andresdelfino:
I have updated the PR.

facundobatista:
Awesome, thank you very much!!!

PR #323, opened 2018-05-22
declare encoding in setup.py

ricardokirkner:


facundobatista:
We do not want the encoding declared in setup.py, it's py3 only, it will only be more confusing

PR #325, opened 2018-06-05
add debian dependencies

gilgamezh:
We were missing this two dependencies

### DIFF
--- a/pkg/debian/control
+++ b/pkg/debian/control
@@ -16,7 +16,7 @@ X-Python3-Version: >= 3.4
 
 Package: fades
 Architecture: all
-Depends: python3-pkg-resources, ${misc:Depends}, ${python3:Depends}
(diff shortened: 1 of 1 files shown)


gilgamezh:
@facundobatista @inakimalerba do we need it? thanks!

inakimalerba:
I don't think you should keep track of the debian/ folder here. It's redundant and useless as it need to be on the root of the repository. The debian packaging is already on salsa [0]

Back to the issue, the dependencies now are set on like this [1]:
```
Depends: ${misc:Depends}, ${python3:Depends},
 python3-pkg-resources,
 python3-distutils,
 ca-certificates,
```
As @facundobatista said on Telegram, if pip is not found, fades installs it inside the venv, which only needed `ca-certificates`.

0_ https://salsa.debian.org/debian/fades/

1_ https://salsa.debian.org/debian/fades/blob/master/debian/control

PR #332, opened 2018-07-09
Split out the dependencies parsing and merging from go() 

andresdelfino:
This addresses #149.


andresdelfino:
Don't know why fades/main.py is reported as a conflicting file. My branch is up to date with the tip of master.

fades/main.py
```diff
@@ -56,13 +57,48 @@
...
 
 
-def _merge_deps(deps):
-    """Merge all the dependencies; latest dicts overwrite first ones."""
-    final = {}
-    for dep in deps:
+def consolidate_dependencies(needs_ipython, child_program,
+                             requirement_files, manual_dependencies):
+    """Parse files, get deps and merge them."""
+    # Deps specified first are overwritten by those specified last
```
facundobatista: This comment is about the generic behaviour of the whole function. It should be part of the docstring.

tests/__init__.py
```diff
@@ -34,3 +36,14 @@ def clean():
     testcase.addCleanup(clean)
 
     return tempfile
+
+
+@nottest
+def generate_test_file(testcase, lines):
+    tempfile = get_tempfile(testcase)
+
+    with open(tempfile, 'w') as f:
```
facundobatista: Don't leave `encoding` unspecified, as it's platform dependent and we don't want that.

tests/__init__.py
```diff
@@ -34,3 +36,14 @@ def clean():
     testcase.addCleanup(clean)
 
     return tempfile
+
+
+@nottest
+def generate_test_file(testcase, lines):
```
facundobatista: Let's keep the previous naming style and call this a "temp" file. Maybe calling the function `create_tempfile`? A nice side effect is that you will not need the ugly `@nottest`

tests/test_main.py
```diff
@@ -41,55 +44,118 @@ def test_prefix_is_not_baseprefix(self):
...
+    """Tests for the gathering stage of consolidate_dependencies."""
+
+    def test_needs_ipython(self):
+        d = main.consolidate_dependencies(needs_ipython=True, child_program=None,
+                                          requirement_files=None, manual_dependencies=None)
+
+        self.assertDictEqual(d, {'pypi': {Requirement.parse('ipython')}})
+
+    def test_child_program(self):
+        child_program = generate_test_file(self, ['"""fades:', 'dep', '"""'])
```
facundobatista: Probably it's easier/cleaner to just use `tests/test_files/req_module.py` here?

fades/main.py
andresdelfino: Yeah, I had some trouble trying to make it all fit in a single < 100 chars line :sweat_smile: Changed it.

tests/__init__.py
andresdelfino: create_tempfile sounds good to me.

tests/test_main.py
andresdelfino: Yeap, didn't look into those test files.

andresdelfino:
@facundobatista requested changes addressed.

facundobatista:
Awesome! Thank you very much!

PR #333, opened 2018-07-22
All gifs showing small functionalities. Fixes #144.

### LINKED ISSUE #144: Create a "terminal video" about using fades
Can be something like it's done here:

  https://github.com/mptre/yank

but I'd prefer a voice in off telling what's going on.

The video should take ~2, 3 minutes.

Also, we could make several small ones, muted, can be even gifs, like "how to use a requirement.txt file", or "how to start a specific django project"... I wouldn't do a gif for _every_ feature, but, say, the top 5....

facundobatista:


gilgamezh:
thanks!

PR #334, opened 2018-08-13
Salute only once and before polluting user's screen. Fixes #301.

### LINKED ISSUE #301: Bad order in messages in stdout
The "Hi user" should be first, always, see the problem here:

```
$ fades -d django==2.0
*** fades ***  2018-03-31 09:39:12,529  INFO     Checking the availabilty of dependencies in PyPI. You can use '--no-precheck-availability' to avoid it.
*** fades ***  2018-03-31 09:39:13,388  INFO     Hi! This is fades 7.0, automatically managing your dependencies
...
```

facundobatista:


fades/logger.py
```diff
@@ -20,11 +20,35 @@
...
+SALUTATION = "Hi! This is fades {}, automatically managing your dependencies".format(__version__)
+
+
+class SalutingStreamHandler(logging.StreamHandler):
+    """A handler that salutes once before polluting user screen.
+
+    Note that the salutation is done in INFO level, to respect "verbose" modifiers.
+    """
+
+    def __init__(self, logger):
```
andresdelfino: Perhaps this new class should take the salutation message as an argument instead of using a module level variable?

facundobatista: You'd be gaining a flexibility that you're not using, with the cost of more complexity. So not worth it...

PR #335, opened 2018-09-17
Support getting the script from the web

facundobatista:
It just supports the script being an URL.


gilgamezh:
+1 we should add it to the docs.

PR #344, opened 2019-02-03
Don't precheck-availability when no dependencies

### LINKED ISSUE #314: Alert of "checking dependencies availability" appears even when no dependencies
```
$ fades
*** fades ***  2018-04-24 15:49:13,147  INFO     Checking the availabilty of dependencies in PyPI. You can use '--no-precheck-availability' to avoid it.
Python 3.5.2 (default, Nov 23 2017, 16:37:01) 
[GCC 5.4.0 20160609] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> 
```

gilgamezh:
fixes #314

### DIFF
--- a/fades/main.py
+++ b/fades/main.py
@@ -299,7 +299,7 @@ def go():
 
     if create_venv:
         # Check if the requested packages exists in pypi.
-        if not args.no_precheck_availability and indicated_deps:
+        if not args.no_precheck_availability and indicated_deps.get('pypi'):
             logger.info("Checking the availabilty of dependencies in PyPI. "
                         "You can use '--no-precheck-availability' to avoid it.")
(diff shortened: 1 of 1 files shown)


fades/main.py
```diff
@@ -299,7 +299,7 @@ def go():
 
     if create_venv:
         # Check if the requested packages exists in pypi.
-        if not args.no_precheck_availability:
+        if not args.no_precheck_availability and indicated_deps:
```
facundobatista: this will pass and still give the message incorrectly if has not-pypi dependencies... you should check kind of doing...
```suggestion
        if not args.no_precheck_availability and indicated_deps.get('pypi'):
```

PR #347, opened 2019-02-03
Check child access

gilgamezh:


tests/test_main.py
```diff
@@ -176,9 +177,20 @@ def test_no_child_at_all(self):
...
+        self.assertEqual(analyzable, child_path)
+        self.assertEqual(child, child_path)
+
+    def test_normal_child_program_not_found(self):
+        with self.assertRaises(FadesError):
+            main.decide_child_program(False, 'does_not_exist.py')
+
+    def test_normal_child_program_no_access(self):
+        child_path = create_tempfile(self, "")
+        os.chmod(child_path, 333)  # Remove read permission.
```
facundobatista: Probably you should change the permission back to something "sane" after the test (make sure it's before the actual file removal from create_tempfile)

fades/main.py
```diff
@@ -96,14 +96,33 @@ def consolidate_dependencies(needs_ipython, child_program,
 
 def decide_child_program(args_executable, args_child_program):
     """Decide which the child program really is (if any)."""
+    # We get the logger here because it's not defined at module level
+    logger = logging.getLogger('fades')
+
     if args_executable:
+        # if --exec given, check that it's just the executable name,
+        # not absolute or relative paths
+        if args_executable and os.path.sep in args_child_program:
```
facundobatista: Double checking!
```suggestion
        if os.path.sep in args_child_program:
```

fades/main.py
```diff
@@ -96,14 +96,33 @@ def consolidate_dependencies(needs_ipython, child_program,
...
+
         # indicated --execute, local and not analyzable for dependencies
         analyzable_child_program = None
         child_program = args_child_program
     elif args_child_program is not None:
         # normal case, the child program is to be analyzed (being it local or remote)
         if args_child_program.startswith(("http://", "https://")):
             args_child_program = helpers.download_remote_script(args_child_program)
+        else:
+            if not os.access(args_child_program, os.R_OK):
```
facundobatista: Shouldn't you be checking for X_OK?

gilgamezh: mmmm we support a non executable file. I think read is ok.

facundobatista: you're right

Issue #359, opened 2019-03-04
Migrate testing infrastructure to pytest


facundobatista:
Because it's the papota.

PR #363, opened 2019-05-09
Changes test_cache, test_logger, test_infra, test_pipmanager to pytest, updated requirements.txt

eduzen:
I think that this task #359 is super big and we can do it in several steps. Pytest allows you to run normal tests with classes and new function tests, so it's not a problem to do it gradually. 

What do you think @facundobatista @gilgamezh  ?


eduzen:
Ok a lot of problems with the version of python. I fixed for all the versions except for python 3.3. 
Can we remove 3.3 for testing?

Also `appveyor` is not working because is trying to use `nosetests` how can I fix that?

eduzen:
@facundobatista we need to configure AppVeyor . It's using nosetest and no pytest. I greped "nose" in my branch and there is nothing. So I assume that is something configured on AppVeyor to run with nose.

PR #374, opened 2019-06-19
Working refactor test_parsing

eduzen:


tests/test_parsing/test_reqs.py
```diff
@@ -0,0 +1,139 @@
+"""Check the requirements parsing."""
+import io
+
+from pkg_resources import parse_requirements
+from fades import parsing, REPO_PYPI, REPO_VCS
+
+
+def get_req(text):
```
facundobatista: Don't use this get_req, but the central one I normalized in the previous branch (that would lead you to remove the "list around the get_reqs call in the rest of the file).

Thanks!

tests/test_parsing/test_file.py
```diff
@@ -0,0 +1,501 @@
...
+    """))
+    assert parsed == {
+        REPO_PYPI: get_reqs('requests')
+    }
+
+
+def test_fades_user_strange_comment_with_hashtag_ignored():
+    parsed = parsing._parse_content(io.StringIO("""
+      import foo # fades==2 # Some comment with #hashtash
+    """))  # noqa
```
facundobatista: May this `# noqa` be removed? Just try it, it looks it's left there from ancient times

PR #377, opened 2019-07-22
Fix use of flake8.get_style_guide

dmascialino:
I found that `test_flake8_pytest` is using .flake8 file instead of `FLAKE8_OPTIONS`.
To reporduce this error you can delete .flake8 file and run `./test`, this will fail with a lot of ` E501 line too long`. 

With this patch I fixed this behavior.


gilgamezh:
thanks!

PR #383, opened 2020-03-28
Comment in the readme about local dependencies.

facundobatista:


gilgamezh:
@facundobatista please rebase after mergin #387

PR #392, opened 2020-05-04
Provide the --freeze parameter. Fixes #385

### LINKED ISSUE #385: Include `--freeze`
Example command: `fades -r requirements.txt --freeze=freezedreqs.txt -x pytest -v -s`

It should `(pip) freeze` the just created venv to the specified file. If the venv is used from the cache, and not re-created, the freeze is a noop.

facundobatista:


README.rst
```diff
@@ -305,6 +305,52 @@ matter which one is the latest. But if you do::
...
+project, you actually get to use a new virtualenv automatically.
+
+If you don't pin the dependencies in your requirements file, this has 
+another nice side effect: everytime you use them in a new environment (or
+if you have `--check-updates` set) you will get latest versions, effectively
+avoiding the trap of sticking in old versions forever.
+
+However, this has a bad side. If it happens that a dependency of your 
+project released a revision between the moment you run the tests and the 
+mement your project is deployed to the server, it may happen that you
```
gilgamezh, line 321: mement

fades/pipmanager.py
```diff
@@ -108,3 +108,10 @@ def _brute_force_install_pip(self):
         python_exe = os.path.join(self.env_bin_path, "python")
         helpers.logged_exec([python_exe, self.pip_installer_fname, '-I'])
         self.pip_installed = True
+
+    def freeze(self, filepath):
+        """Dump venv contents to the indicated filepath."""
+        logger.debug("running freeze to store in %r", filepath)
+        stdout = helpers.logged_exec([self.pip_exe, "freeze", "--all", "--local"])
+        with open(filepath, "wt", encoding='utf8') as fh:
```
gilgamezh, line 116: why are you adding `t` here? That's just the default

README.rst
facundobatista: Gotcha

fades/pipmanager.py
facundobatista: Explicit better than implicit :).

Now seriously, I always write 't' or 'b' there, as the difference is getting bytes or unicode chars, which may be critical in the following handling of the file handler.

PR #397, opened 2020-05-19
Get pip automatically upgraded to latest version (unless explicitly avoided)

facundobatista:


tests/test_envbuilder.py
```diff
@@ -88,21 +89,26 @@ def test_create_simple(self):
                 'dep2': 'v2',
             }
         })
+        expected_pipmanager_call = call(
+            'env_bin_path', pip_installed='pip_installed', options=[],
+            avoid_pip_upgrade=avoid_pip_upgrade)
+        self.assertEqual(mock_mgr_c.call_args, expected_pipmanager_call)
```
gilgamezh, line 95: you can always use `assert_called_once_with(*args, **kwargs)`  https://docs.python.org/3/library/unittest.mock.html#unittest.mock.Mock.assert_called_once_with

PR #415, opened 2022-03-22
Fixes to make fades to actually work in a core20 snap.

facundobatista:


gilgamezh:
LGTM! 
I don't understand why "some checks were not successful"

PR #420, opened 2023-11-23
chore: update py3 support list

chenrui333:
- https://github.com/Homebrew/homebrew-core/pull/149531


facundobatista (approved):
Great, thanks!

PR #421, opened 2023-11-23
chore: upgrade action deps

chenrui333:
- upgrade actions to base on nodejs 20 runtime

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -18,12 +18,10 @@ jobs:
 
     runs-on: ${{ matrix.os }}
     steps:
-      - name: Checkout code
-        uses: actions/checkout@v2
-        with:
-          fetch-depth: 0
+      - uses: actions/checkout@v4
+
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yaml
```diff
@@ -18,12 +18,10 @@ jobs:
 
     runs-on: ${{ matrix.os }}
     steps:
-      - name: Checkout code
-        uses: actions/checkout@v2
-        with:
-          fetch-depth: 0
```
facundobatista, line 24: Hello! Thanks for this! Question, why the removal of `fetch-depth`? Thanks!

chenrui333: Removing the fetch-depth would just do the HEAD build, would it disrupt anything?

facundobatista: Ah, I see

facundobatista (approved):
Thanks!!

PR #433, opened 2024-06-11
Only use the `venv` module to create virtualenvs.

facundobatista:
This is possible now as we're living in the future already (?): Python minimum version is now 3.6, which is old enough.

While it simplifies the code a lot, there may be some breakage. 

If the user is passing specific options to `virtualenv`, as we don't use it anymore, what to do? For that I added the possibility of passing options to the `venv` module itself... the user may need to change something in their scripts, but at least the can keep things in control.

Also I improved the README, stopped using "virtualenv" to mean "virtual environment" as it was too confusing. Being explicit now.

### DIFF
--- a/.github/workflows/integtests.yaml
+++ b/.github/workflows/integtests.yaml
@@ -78,6 +78,7 @@ jobs:
 
       - name: Also set up Python 3.10 for cross-Python test
         uses: actions/setup-python@v5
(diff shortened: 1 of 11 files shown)


gilgamezh (approved):
LGTM! did some extra checks locally and works as expected

PR #440, opened 2024-07-18
Update different details now we support Py>-3.6 only

### LINKED ISSUE #429: Cleanups after we support only Python 3.6 or greater
When #411 lands, we can improve code:

- Stop declaring constants in fades/__init__.py:

```
# Not using http.server.HTTPStatus to support python < 3.5
HTTP_STATUS_NOT_FOUND = 404
HTTP_STATUS_OK = 200
```

- There are cleanups in `tests/conftest.py` (see comments there).

facundobatista:
Specifically:
- stop defining http status codes ourselves (they are in `http.server.HTTPStatus`)
- stop using `datetime.utcnow()` (in favor of using `datetime.now(UTC)`)
- simplified temp files fixtures

Fixes #429


gilgamezh (approved):
Nice!

PR #441, opened 2025-06-06
Show an error if the `packaging` dependency is not present

alejandrodau:
MacOS lacks the `packaging` package by default. So running `bin/fades -V` fails with a stack trace.

This change is to show a more readable error when the required package is missing.


facundobatista:
Thanks!

PR #452, opened 2026-06-18
Add PEP 723 inline script metadata support

### LINKED ISSUE #423: PEP 723 support
[PEP 723](https://peps.python.org/pep-0723/) introduces a standard way to declare dependencies in single-file scripts. It has been accepted and has now been implemented in [pip-run](https://github.com/jaraco/pip-run) and [pipx](https://github.com/pypa/pipx), which run scripts with dependencies like fades. I think fades should support it, too; then fades could run other runners' scripts and vice versa.

Here is an example of a script with embedded dependency data from the PEP:

```python
# /// script
# requires-python = ">=3.11"
# dependencies = [
#   "requests<3",
#   "rich",
# ]
# ///

import requests
from rich.pretty import pprint

resp = requests.get("https://peps.python.org/api/peps.json")
data = resp.json()
pprint([(k, v["title"]) for k, v in data.items()][:10])
```

gilgamezh:
Closes #423 

[PEP 723](https://peps.python.org/pep-0723/) defines a standard `# /// script` comment block (TOML) for declaring a single-file script's `dependencies` and `requires-python`. It's implemented by pipx, pip-run and uv. This adds support to fades so it can run scripts written for other runners — and so scripts written for fades can run on them.

## What it does

- **`dependencies`**: parsed from the block and merged like every other dependency source (inline `# fades` marks, docstrings, `-r` files, `-d` flags).
- **`requires-python`**: honored by selecting a suitable interpreter — the currently selected one (`--python` or fades' own) if it satisfies the specifier, otherwise an auto-discovered `pythonX.Y` on `PATH`, failing cleanly if none is available. An explicit `--python` that conflicts is reported rather than silently overridden, so the user stays in control.
- **TOML parsing**: stdlib `tomllib` on Python 3.11+, falling back to `tomli` on older versions (conditional dependency in `setup.py`).
- **Robust errors**: malformed metadata (bad TOML, invalid requirement string, bad/non-string `requires-python`, non-list `dependencies`, multiple `script` blocks) raises a clean `FadesError` with an explanatory log line instead of dumping a traceback.

Example:

```python
# /// script
# requires-python = ">=3.11"
# dependencies = [
#   "requests<3",
#   "rich",
# ]
# ///

import requests
from rich.pretty import pprint
```

## ⚠️ Point for review: `requires-python` behavior

This goes **beyond pipx and pip-run, which both ignore `requires-python`** (only uv honors it, via its own interpreter management). Since fades has no interpreter management, the chosen behavior is:

- selected interpreter satisfies the specifier → use it;
- it doesn't, and no explicit `--python` → discover a matching `pythonX.Y` on `PATH`, else fail;
- it doesn't, and `--python` *was* given → fail (don't override the user's explicit choice).

Discovery is best-effort over `PATH` (`python3.6`–`python3.29`, `python3`, `python`). Happy to dial this back to validate-only if preferred.

## Tests

New `tests/test_parsing/test_pep723.py` plus fixtures, interpreter-selection/version-probe tests in `test_helpers.py`, and a `consolidate_dependencies` test in `test_main.py`. Full suite green (`./test`), flake8 clean. Docs updated in `README.rst` and `man/fades.1`.

### DIFF
--- a/fades/helpers.py
+++ b/fades/helpers.py
@@ -310,9 +310,12 @@ def check_pypi_updates(dependencies):
 
 def _pypi_head_package(dependency):
     """Hit pypi with a http HEAD to check if pkg_name exists."""
-    if dependency.specifier:
-        spec = list(dependency.specifier)[0]
-        version = spec.version
+    # Only an exact pin (== or ===) maps to a version-specific URL; range specifiers
+    # like '<3' or '>=2' must be checked by package name, otherwise we'd query a
+    # non-existent "version" and wrongly conclude the package is missing.
+    exact_specs = [spec for spec in dependency.specifier if spec.operator in ("==", "===")]
+    if exact_specs:
+        version = exact_specs[0].version
         url = BASE_PYPI_URL_WITH_VERSION.format(name=dependency.name, version=version)
     else:
         url = BASE_PYPI_URL.format(name=dependency.name)
--- a/fades/helpers.py
+++ b/fades/helpers.py
@@ -17,6 +17,7 @@
 """A collection of utilities for fades."""
 
 import json
+import shutil
 import logging
 import os
 import subprocess
@@ -28,11 +29,16 @@ from urllib import request, parse
 from urllib.error import HTTPError
 
 from packaging.requirements import Requirement
+from packaging.specifiers import InvalidSpecifier, SpecifierSet
 from packaging.version import Version
 
 from fades import FadesError, _version
 
 logger = logging.getLogger(__name__)
+
+# range of CPython 3 minor versions probed when auto-selecting an interpreter to
+# satisfy a PEP 723 'requires-python' specifier
+PYTHON_MINOR_RANGE = range(6, 30)
 
 # command to retrieve the version from an external Python
 SHOW_VERSION_CMD = """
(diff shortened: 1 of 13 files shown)


gilgamezh:
Did run tests here because we only run them when target is master 

https://github.com/PyAr/fades/actions/runs/27750679594

should be executed when we change the target branch

fades/helpers.py
```diff
@@ -162,6 +168,93 @@ def get_interpreter_version(requested_interpreter):
...
+        raise FadesError("Could not get interpreter version")
+
+
+def _find_interpreter(specifier):
+    """Search PATH for a python interpreter whose version satisfies the specifier."""
+    candidate_names = ["python3.{}".format(minor) for minor in PYTHON_MINOR_RANGE]
+    candidate_names += ["python3", "python"]
+
+    found = {}  # path -> Version, to avoid probing the same interpreter twice
+    for name in candidate_names:
```
facundobatista, line 194: It feels a little clumsy to run an external subprocess 25 times just in case to find something.

What about exploring the PATH to see what's out there? See [this code](https://linkode.org/#daxRBOTySL0sYQS70X0Xe5), for example, in my machine it produces:
```
('/usr/bin/python3.12', '3.12.3')
('/usr/bin/python3.14', '3.14.6')
('/usr/bin/python3.13', '3.13.14')
('/usr/bin/python3.14t', '3.14.6')
('/usr/bin/python3.11', '3.11.15')
```

We can work from there...

fades/helpers.py
```diff
@@ -162,6 +168,93 @@ def get_interpreter_version(requested_interpreter):
...
+
+    if requested_python is not None:
+        # the user explicitly chose an interpreter that conflicts with the script; don't
+        # silently override that choice, fail so they can resolve it
+        msg = ("The chosen Python interpreter (version {}) does not satisfy the script's "
+               "requires-python ({!r})".format(current_version, requires_python))
+        logger.error(msg)
+        raise FadesError(msg)
+
+    # nothing was explicitly requested and fades' own python doesn't satisfy the spec:
```
facundobatista, line 245: this is a little confusing... if nothing was explicitly requested, what is "the spec"?

fades/helpers.py
```diff
@@ -217,9 +310,12 @@ def check_pypi_updates(dependencies):
...
     """Hit pypi with a http HEAD to check if pkg_name exists."""
-    if dependency.specifier:
-        spec = list(dependency.specifier)[0]
-        version = spec.version
+    # Only an exact pin (== or ===) maps to a version-specific URL; range specifiers
+    # like '<3' or '>=2' must be checked by package name, otherwise we'd query a
+    # non-existent "version" and wrongly conclude the package is missing.
+    exact_specs = [spec for spec in dependency.specifier if spec.operator in ("==", "===")]
+    if exact_specs:
+        version = exact_specs[0].version
```
facundobatista, line 318: This may be a bug we have... but this code is not the solution at all.

The bug I think we have: "if you declare a version X not using ==, this will check for X anyway"

The bug this is introducing: "if you declare a version X not using ==, this will just check if the package exists (not caring the version)".

You should open a GH issue (if not there already) and let's handle separately,

fades/main.py
```diff
@@ -388,8 +391,13 @@ def go():
     if args.check_updates:
         helpers.check_pypi_updates(indicated_deps)
 
+    # honor a PEP 723 'requires-python' (this may pick a different interpreter than the one
+    # explicitly requested with --python or fades' own)
+    _, requires_python = parsing.parse_pep723(analyzable_child_program)
```
facundobatista, line 396: Can we avoid the duplicate parsing?

tests/test_helpers.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2015-2026 Facundo Batista, Nicolás Demarchi
+# Copyright 2014-2026 Facundo Batista, Nicolás Demarchi
```
facundobatista, line 1: Why are you changing the "from" in this copyright?

tests/test_files/pep723_and_comment.py
```diff
@@ -0,0 +1,13 @@
+# Copyright 2014-2026 Facundo Batista, Nicolás Demarchi
```
facundobatista, line 1: New files (here and all below) should start with copyright in current year. And they're missing the rest of the header.

fades/parsing.py
```diff
@@ -304,3 +317,64 @@ def parse_docstring(filepath: Path):
...
+    dependencies = metadata.get('dependencies', [])
+    if not isinstance(dependencies, list):
+        logger.error(
+            "PEP 723 'dependencies' must be a list, got %s", type(dependencies).__name__)
+        raise FadesError("PEP 723 'dependencies' must be a list")
+    if dependencies:
+        # PEP 723 dependencies are standard PEP 508 strings (including 'name @ url' direct
+        # references that pip understands), so they all go to the PyPI repo.
+        try:
+            deps[REPO_PYPI] = [Requirement(dep) for dep in dependencies]
```
facundobatista, line 364: Dependencies from other backends should not go in `REPO_PYPI`, and should not be wrapped in a `Requirement` (because there are glitches in the comparison later, that's why we use VCSDependency and such) ... take a look at `parse_fade_requirement`

fades/helpers.py
```diff
@@ -162,6 +168,93 @@ def get_interpreter_version(requested_interpreter):
...
+    candidates = sorted(
+        (version, path) for path, version in found.items()
+        if specifier.contains(version, prereleases=True))
+    if not candidates:
+        return None
+    # pick the highest satisfying version
+    return candidates[-1][1]
+
+
+def get_interpreter_for_requirement(requires_python, requested_python):
```
facundobatista, line 213: These names are very confusing. I know `requires_python` is PEP723 naming, but both names almost mean the same.

What about being explicit and call them `python_toml` and `python_args` or similar?

facundobatista (changes requested):
Great work, thanks for this! I annotated some details and comments for stuff we need to think about...

PR #454, opened 2026-06-20
Add optional uv backend (opt-in via --use-uv / config)

### LINKED ISSUE #451: Proposal: support uv as an optional backend for venv creation & installs
## Context

Relates to #423 (PEP 723 support).

[`uv`](https://docs.astral.sh/uv/) has become the de-facto fast Python package/venv manager and overlaps heavily with what fades does. This issue does two things:

1. Documents a **feature-by-feature comparison** of every fades command/feature against its `uv` equivalent.
2. Proposes a concrete first step: **use `uv` as an optional backend** (create the venv + install deps with `uv` instead of `pip`/`venv` when `uv` is available), keeping all the rest of fades' behaviour unchanged.

The comparison is meant to inform a *future* decision about fades' direction. For now the only proposed action is the optional-backend work below.

## fades → uv feature comparison

Docs read: uv latest ([scripts](https://docs.astral.sh/uv/guides/scripts/), [run](https://docs.astral.sh/uv/concepts/projects/run/), [CLI reference](https://docs.astral.sh/uv/reference/cli/)) and fades `master` README/CLI.

### CLI options

| fades option | What it does | uv equivalent | Same? |
|---|---|---|---|
| `fades script.py` | Run script in managed venv | `uv run script.py` | ✅ |
| `#!/usr/bin/env fades` | Self-executing script | `#!/usr/bin/env -S uv run --script` | ✅ |
| `fades http://.../script.py` | Run script from pastebin/URL | `uv run https://.../script.py` | ✅ (fades also resolves pastebin/gist/linkode share pages) |
| `import x # fades` | Mark dep via import comment | ❌ none — uv uses PEP 723 or `--with` | ❌ fades-only |
| docstring deps | Deps in docstring after `fades` line | ❌ none | ❌ fades-only |
| `-d, --dependency` | Add dep on CLI (repeatable) | `uv run --with PKG` | ✅ |
| `-r, --requirement` | Install from requirements file(s) | `uv run --with-requirements req.txt` | ✅ |
| `-p, --python` | Pick interpreter | `uv run --python 3.11` (also auto-downloads) | ✅ (uv superset) |
| `-i, --ipython` | Use IPython REPL | `uv run --with ipython ipython` | ⚠️ manual, not a flag |
| `-a, --autoimport` | Auto-import deps in REPL | ❌ none | ❌ fades-only |
| (no script) | Open REPL in the env | `uv run python` | ✅ |
| `--system-site-packages` | Venv sees system packages | `uv venv --system-site-packages` (create-time only) | ⚠️ partial |
| `-x, --exec` | Run any installed executable in venv | `uv run <exe>` / `uvx <exe>` | ✅ |
| `-m, --module` | Run module | `uv run -m module` | ✅ |
| `--check-updates` / `-U` | Re-resolve to newest | `uv run --refresh` / `--upgrade` | ✅ |
| `--freeze FILE` | Dump `pip freeze` of env | `uv lock --script` (`.py.lock`) / `uv export` | ⚠️ different format |
| `--no-precheck-availability` | Skip PyPI existence check | n/a (uv resolves directly) | — |
| `--avoid-pip-upgrade` | Don't auto-upgrade pip | n/a (no pip) | — |
| `--pip-options` | Pass-through pip args | native flags only; no generic passthrough | ⚠️ partial |
| `--python-options` | Pass-through python args | `uv run python <opts>` | ✅ |
| `--virtualenv-options` | Pass-through virtualenv args | limited `uv venv` flags | ⚠️ partial |
| `-v/--verbose`, `-q/--quiet` | Verbosity | `uv -v` / `uv -q` | ✅ |
| `-V, --version` | Version + system info | `uv version` | ✅ |

### Repository / source support

| fades feature | uv equivalent | Same? |
|---|---|---|
| `pypi::pkg` | default PyPI | ✅ |
| `vcs::git+https://...` | `--with "pkg @ git+https://..."` | ✅ |
| `file:///local/project` | `--with /local/path` | ✅ |
| alternate index | `uv run --index URL` | ✅ |

### Venv lifecycle / maintenance

| fades feature | uv equivalent | Same? |
|---|---|---|
| `--get-venv-dir` / `--where` | ❌ envs live opaquely in uv cache | ❌ fades-only |
| `--rm UUID` | `uv cache clean` / `prune` (no per-UUID removal) | ⚠️ partial |
| `--clean-unused-venvs=N` | `uv cache prune` (no day-based policy) | ⚠️ partial |
| per-depset venv reuse | automatic (keyed on resolved depset) | ✅ |

### Configuration

| fades feature | uv equivalent | Same? |
|---|---|---|
| `.ini` at system/user/project, merged | `uv.toml` / `[tool.uv]` + env vars, layered | ✅ (different format) |
| config-injected global deps into every env | ❌ none | ❌ fades-only |

### Genuine fades-only features (uv can't replicate today)

1. `# fades` import-comment & docstring dependency marking (fades' signature feature).
2. `--autoimport` into the REPL.
3. `--where`/`--get-venv-dir` and `--rm UUID` (per-venv management by UUID).
4. `--clean-unused-venvs=N` day-based GC.
5. Config-injected global dependencies.

## How the comparison shifts once fades supports PEP 723 (#423)

Once #423 lands, fades will read `# /// script ... ///` blocks. This narrows the gap meaningfully:

- **Interoperability**: fades and uv (and pipx/pip-run) could run each other's scripts. The `# fades` comment convention stays as the fades-native way; PEP 723 becomes the portable way.
- **Dependency declaration** moves from "fades-only" toward "✅ same standard" — the biggest semantic difference between the tools largely disappears.
- What remains distinctly fades: the `# fades` comment style, `--autoimport`, explicit per-UUID venv management (`--where`/`--rm`), day-based venv GC, and config-injected global deps.

So post-#423 the two tools converge on the *input format* but still differ on *venv management philosophy* (fades exposes/manages individual venvs by UUID; uv hides them in a cache).

## Proposal: use `uv` as an optional backend

Add `uv` as an **optional** dependency. When `uv` is present on the system, fades uses it for the parts where it's strictly better; when it's absent, fades behaves exactly as today.

Concretely:

- **Venv creation**: use `uv venv` instead of the `venv`/`virtualenv` path.
- **Dependency installation**: use `uv pip install` (or `uv add`) instead of `pip`.
- **Everything else stays fades**: dependency parsing (`# fades` comments, docstring, `-d`/`-r`), venv indexing by UUID, reuse logic, `--where`/`--rm`/`--clean-unused-venvs`, config files, REPL/`--autoimport`, etc.

This gives fades users uv's speed and built-in Python provisioning without changing fades' UX or losing any of its unique features, and it's a low-risk, incremental step.

Open questions to resolve before implementing:

- Detection & opt-out: auto-detect `uv` on `PATH`, with a config flag / env var to force on/off?
- Does `uv`'s venv layout stay compatible with fades' UUID indexing and reuse-by-hash logic?
- How to map `--pip-options` / `--virtualenv-options` onto `uv` invocations.
- `--freeze` semantics when using `uv`.

## Scope of this issue

This issue is for **documenting the comparison and discussing the optional-backend approach**. No implementation is proposed here yet beyond the outline above — the broader question of fades' long-term direction vs. uv is intentionally left for a future decision.

gilgamezh:
## What

Adds [`uv`](https://github.com/astral-sh/uv) as an **opt-in** backend for venv creation and dependency installs. By default fades keeps using `venv`/`pip`; uv is used only when explicitly requested. Everything else (dependency parsing, venv indexing/cache, GC, `--where`/`--rm`, REPL/autoimport, config files) is unchanged.

Closes #451.

## Enabling uv

- **`--use-uv`** — use uv for this run.
- **`--uv-path PATH`** — point at a specific uv executable (implies `--use-uv`); otherwise uv is looked up in `PATH` and validated. If uv is requested but not found, fades **errors out**.
- **`use_uv=true`** in any fades config file (`/etc/fades/fades.ini`, the per-user `fades.ini`, or a repo-local `./.fades.ini`) enables it permanently; a CLI flag still wins.

## Design notes

- **Opt-in, not auto-detect** (per discussion in #451): uv produces different venv contents than pip, so fades' behaviour shouldn't change merely because uv is installed.
- uv has no Python API (astral-sh/uv#6080 closed without one), so it's invoked via subprocess like the existing `PipManager`. fades does **not** depend on the `uv` PyPI package (it runs against the system Python); it uses a uv **binary** on PATH.
- `UvManager` mirrors `PipManager`'s `install`/`get_version`, dispatched from `envbuilder.create_venv`. Backend resolution + option validation live in `main._resolve_uv_backend` (unit-tested).
- `uv venv --seed` so the venv ships pip (and setuptools, per Python version) like the classic backend.
- When uv is used: the PyPI availability pre-check is skipped (uv resolves directly); `--pip-options`/`--venv-options` are errors (use **`--uv-pip-options`**); `--freeze` uses plain pip freeze on the seeded venv.

## Tests / CI

- Unit tests for `_resolve_uv_backend` (enabled/disabled, `--uv-path`, not-found, option conflicts), `UvManager`, envbuilder dispatch, `get_uv_exe`.
- Integration jobs (archlinux/fedora/native-windows) exercise: default = pip, `--use-uv` = uv, conflict + bogus-path errors, and `--freeze` on a uv venv. Windows confirms the uv path end-to-end.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -394,21 +394,37 @@ Examples:
 Using uv as backend
 -------------------
 
-If a `uv <https://github.com/astral-sh/uv>`_ binary is found in ``PATH``, fades automatically uses it as a faster backend to create the virtual environment and install the dependencies (instead of ``venv`` and ``pip``). Nothing else changes: fades keeps doing its own dependency parsing, venv indexing, garbage collection, etc.
+fades can optionally use `uv <https://github.com/astral-sh/uv>`_ as a (much faster) backend to create the virtual environment and install the dependencies, instead of ``venv`` and ``pip``. This is **opt-in**: fades does not change its behaviour just because ``uv`` happens to be installed, because ``uv`` produces slightly different virtual environments than ``pip``.
 
-uv is detected as a binary in ``PATH`` (installed via your system package manager, the standalone installer, etc.); fades does not depend on the ``uv`` PyPI package.
+Enable it per run with ``--use-uv``::
 
-fades runs ``uv venv --seed``, so the virtual environment ships ``pip`` and ``setuptools`` just like the classic ``venv`` backend, keeping packages that expect them present working.
+    fades --use-uv -d requests -x python
 
-If you want to force the classic ``pip`` backend even when uv is available, use ``--no-uv``:
+``uv`` must be available: fades looks it up in ``PATH`` (installed via your system package manager, the standalone installer, etc.; fades does not depend on the ``uv`` PyPI package). You can point at a specific executable with ``--uv-path /path/to/uv`` (which also implies ``--use-uv``). If ``uv`` is requested but can't be found, fades errors out.
 
-``fades --no-uv -d requests -x python``
+Everything else stays fades: dependency parsing (``# fades`` comments, docstring, ``-d``/``-r``), venv indexing and reuse, ``--where``/``--rm``/``--clean-unused-venvs``, config files, the REPL and ``--autoimport``, etc.
+
+Some notes when using ``uv``:
+
+- fades runs ``uv venv --seed``, so the virtual environment ships ``pip`` (and ``setuptools``, depending on the Python version) just like the classic ``venv`` backend, keeping packages that expect them present working.
+- The PyPI availability pre-check is skipped (``uv`` resolves directly and fails early on its own).
+- ``--pip-options`` and ``--venv-options`` are **not** valid together with ``--use-uv`` (they target ``pip``/``venv``); use ``--uv-pip-options`` to pass extra options to ``uv pip install`` instead.
+
+Enabling uv permanently (via config)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+If you always want the uv backend, set ``use_uv=true`` under the ``[fades]`` section of any fades config file. fades reads (in increasing priority) ``/etc/fades/fades.ini`` (machine-wide), the per-user config (``<xdg-config>/fades/fades.ini``) and a repo-local ``./.fades.ini`` (handy to enable uv for a single project). A ``--use-uv``/``--no``-style flag on the command line still wins over the config.
+
+To turn it on for your user without editing files by hand, run::
+
+    python -c "import configparser; from fades.helpers import get_confdir; \
+    p = get_confdir() / 'fades.ini'; c = configparser.ConfigParser(); c.read(p); \
+    c.has_section('fades') or c.add_section('fades'); c.set('fades', 'use_uv', 'true'); \
+    c.write(p.open('w')); print('uv enabled by default in', p)"
 
 A note about ``pkg_resources``: some packages (e.g. ``azure-cli``) still import the long-deprecated ``pkg_resources``, which was **removed** in ``setuptools >= 81``. uv (and the classic backend on Python 3.12+) installs a recent ``setuptools`` that no longer provides it, so those imports fail. This is not specific to the uv backend; if you hit it, add an older setuptools as an explicit dependency::
 
-    fades -d azure-cli -d "setuptools<81" -x az --help
-
-(or use ``--no-uv`` on a Python version old enough to seed an old setuptools).
+    fades --use-uv -d azure-cli -d "setuptools<81" -x az --help
 
 
 Setting options using config files
--- a/README.rst
+++ b/README.rst
@@ -417,8 +417,8 @@ If you always want the uv backend, set ``use_uv=true`` under the ``[fades]`` sec
 
 To turn it on for your user without editing files by hand, run::
 
-    python -c "import configparser; from fades.helpers import get_confdir; \
-    p = get_confdir() / 'fades.ini'; c = configparser.ConfigParser(); c.read(p); \
+    python -W ignore -c "import configparser; from pathlib import Path; from fades.helpers import get_confdir; \
+    p = Path(get_confdir()) / 'fades.ini'; c = configparser.ConfigParser(); c.read(p); \
     c.has_section('fades') or c.add_section('fades'); c.set('fades', 'use_uv', 'true'); \
     c.write(p.open('w')); print('uv enabled by default in', p)"
 
--- a/README.rst
+++ b/README.rst
@@ -391,6 +391,18 @@ Examples:
 ``fades --python-options=-B foo.py``
 
 
+Using uv as backend
+-------------------
+
+If a `uv <https://github.com/astral-sh/uv>`_ binary is found in ``PATH``, fades automatically uses it as a faster backend to create the virtual environment and install the dependencies (instead of ``venv`` and ``pip``). Nothing else changes: fades keeps doing its own dependency parsing, venv indexing, garbage collection, etc.
+
+uv is detected as a binary in ``PATH`` (installed via your system package manager, the standalone installer, etc.); fades does not depend on the ``uv`` PyPI package.
+
+If you want to force the classic ``pip`` backend even when uv is available, use ``--no-uv``:
+
+``fades --no-uv -d requests -x python``
+
+
 Setting options using config files
 ----------------------------------
 
--- a/README.rst
+++ b/README.rst
@@ -398,9 +398,17 @@ If a `uv <https://github.com/astral-sh/uv>`_ binary is found in ``PATH``, fades 
 
 uv is detected as a binary in ``PATH`` (installed via your system package manager, the standalone installer, etc.); fades does not depend on the ``uv`` PyPI package.
 
+fades runs ``uv venv --seed``, so the virtual environment ships ``pip`` and ``setuptools`` just like the classic ``venv`` backend, keeping packages that expect them present working.
+
 If you want to force the classic ``pip`` backend even when uv is available, use ``--no-uv``:
 
 ``fades --no-uv -d requests -x python``
+
+A note about ``pkg_resources``: some packages (e.g. ``azure-cli``) still import the long-deprecated ``pkg_resources``, which was **removed** in ``setuptools >= 81``. uv (and the classic backend on Python 3.12+) installs a recent ``setuptools`` that no longer provides it, so those imports fail. This is not specific to the uv backend; if you hit it, add an older setuptools as an explicit dependency::
+
+    fades -d azure-cli -d "setuptools<81" -x az --help
+
+(or use ``--no-uv`` on a Python version old enough to seed an old setuptools).
 
 
 Setting options using config files
--- a/README.rst
+++ b/README.rst
@@ -491,9 +491,9 @@ Execute the Python interactive interpreter in a virtual environment after instal
 
     fades -r requirements.txt -r requirements_devel.txt
 
-Use the ``django-admin.py`` script to start a new project named ``foo``, without having to have django previously installed::
+Use the ``django-admin`` script to start a new project named ``foo``, without having to have django previously installed::
 
-    fades -d django -x django-admin.py startproject foo
+    fades -d django -x django-admin startproject foo
 
 Remove a virtual environment matching the given uuid from disk and cache index::
 
--- a/fades/envbuilder.py
+++ b/fades/envbuilder.py
@@ -19,6 +19,7 @@
 import logging
 import os
 import shutil
+import sys
 from datetime import datetime, timezone
 from pathlib import Path
 from uuid import uuid4
@@ -27,6 +28,7 @@ from venv import EnvBuilder
 from fades import FadesError, REPO_PYPI, REPO_VCS
 from fades import helpers, cache
 from fades.pipmanager import PipManager
+from fades.uvmanager import UvManager
 from fades.multiplatform import filelock
 
 logger = logging.getLogger(__name__)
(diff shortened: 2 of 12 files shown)


gilgamezh:
### Benchmark: uv vs pip

Wall time for `fades -d <pkg> -x <run-something-in-the-venv>`, comparing the default (uv) backend against `--no-uv` (pip).

#### `django` (single dependency)

| Scenario | Backend | Wall time |
|---|---|---|
| Fresh venv (create + install) | pip (`--no-uv`) | **3.43s** |
| Fresh venv (create + install) | uv | **0.39s** |
| Existing/cached venv (no install) | either | ~0.16s |

≈ **8.7× faster** on the create+install path.

#### `jupyterlab` (~60 transitive deps)

| Scenario | Backend | Wall time |
|---|---|---|
| Fresh, **cold** package cache (incl. downloads) | pip (`--no-uv`) | **16.63s** |
| Fresh, **cold** package cache (incl. downloads) | uv | **12.70s** |
| Fresh, **warm** package cache (already downloaded) | pip (`--no-uv`) | **13.97s** |
| Fresh, **warm** package cache (already downloaded) | uv | **1.62s** |
| Existing/cached venv (no install) | either | **0.92s** |

- **Cold cache** (first-ever install): ≈ **1.3×**. On a first install both backends spend most of the time downloading the same ~60 wheels over the network, which dominates and masks the backend difference.
- **Warm cache** (steady state, packages already fetched once): **13.97s → 1.62s ≈ 8.6× faster** — this isolates uv's actual resolve+install speed and matches the django ratio.

#### Takeaways

- uv is **~8–9× faster on the install itself**; the small cold-cache jupyterlab number is just because raw download bandwidth (identical for both backends) dominates the very first run.
- The **cached-venv** path is unchanged regardless of backend — no install happens, fades just reuses the venv.

_Single run per row, warm/cold as noted; numbers will vary with network and machine._

gilgamezh:
### uv venvs now seed pip/setuptools (`uv venv --seed`) + a note on `pkg_resources`

While testing `azure-cli` we hit `ModuleNotFoundError: No module named 'pkg_resources'` on the uv backend, which surfaced two separate things:

**1. uv venvs were missing pip/setuptools (fixed)**

`uv venv` seeds nothing by default, whereas the classic `python -m venv` backend ships `pip` (and, on Python ≤ 3.11, `setuptools`). So uv-built venvs silently lacked tooling that some packages expect to be present. fades now runs **`uv venv --seed`**, restoring that parity:

- Python 3.12+: seeds `pip` (matches stdlib `venv`, which dropped setuptools seeding in 3.12).
- Python ≤ 3.11: seeds `pip` + `setuptools`.

Cost is negligible (~6 ms, warm cache).

**2. `pkg_resources` / `azure-cli` (not fixable transparently — documented)**

`azure-cli` imports the long-deprecated `pkg_resources`, which **setuptools removed in `>= 81`**. `--seed` installs a recent setuptools, so it does **not** bring `pkg_resources` back. Importantly, this is **not uv-specific** — the pip backend fails the same way on Python 3.12+ (no setuptools seeded at all). It only "works" on the classic backend on Python ≤ 3.11 because that seeds an *old* bundled setuptools that still ships `pkg_resources`.

fades shouldn't force-pin an old setuptools, so the workaround is documented in the README — request an older setuptools explicitly:

```
fades -d azure-cli -d "setuptools<81" -x az --help
```

(or `--no-uv` on a Python old enough to seed an old setuptools). Verified this restores `pkg_resources` (setuptools 80.10.2) through the uv backend.

Implemented in 9ab2790.

gilgamezh:
### Switched uv from default to **opt-in**

Following the discussion in #451 (thanks @facundobatista), uv is no longer auto-detected/used when present. It's now explicit:

- `--use-uv` (or `--uv-path`) per run, or `use_uv=true` in a config file (machine/user/repo `.fades.ini`) to enable it always — CLI overrides config.
- requesting uv when it isn't installed is an **error** (no silent fallback).
- `--pip-options`/`--venv-options` are rejected with uv; use the new `--uv-pip-options` for `uv pip install` passthrough.
- the PyPI pre-check is skipped under uv; `--freeze` uses standard pip-freeze output (venvs are seeded with pip).

The `FADES_USE_UV` env var idea was dropped in favor of config files, which already support per-repo/per-machine scoping plus CLI override. The README documents a one-liner to flip `use_uv=true` into your `fades.ini`.

All green on Linux/macOS/Windows.

fades/envbuilder.py
facundobatista, line 1: Remember updating copyright years.

fades/main.py
```diff
@@ -281,6 +308,18 @@ def go():
...
+        help="use uv (instead of venv/pip) to create the virtualenv and install dependencies; "
+             "uv must be available in PATH or indicated with --uv-path.")
+    parser.add_argument(
+        '--uv-path', action='store', metavar='UV_PATH',
+        help="path to the uv executable to use (implies --use-uv); if not given, uv is looked up "
+             "in PATH.")
+    parser.add_argument(
+        '--uv-pip-options', action='append', default=[],
+        help="extra options to be supplied to 'uv pip install' (this option can be used multiple "
+             "times); only valid together with --use-uv.")
```
facundobatista, line 322: Please rephrase this (as it also may be used if --uv-path is included). Maybe saying something like "only valid if 'uv' is indicated to use"?

fades/uvmanager.py
```diff
@@ -0,0 +1,86 @@
+# Copyright 2024-2026 Facundo Batista, Nicolás Demarchi
```
facundobatista, line 1: Only '2026' here.

man/fades.1
```diff
@@ -142,6 +145,18 @@ Run library module as a script (same behaviour than Python's \fB-m\fR parameter)
...
+.BR --use-uv
+Use \fBuv\fR (instead of \fBvenv\fR/\fBpip\fR) as the backend to create the virtual environment and install dependencies. This is opt-in: fades uses \fBpip\fR by default even when \fBuv\fR is installed. \fBuv\fR must be available in \fBPATH\fR (or indicated with \fB--uv-path\fR), otherwise fades errors out. It can also be enabled permanently with \fBuse_uv=true\fR in a config file. Not compatible with \fB--pip-options\fR/\fB--venv-options\fR.
+
+.TP
+.BR --uv-path =\fIUV_PATH\fR
+Path to the \fBuv\fR executable to use (implies \fB--use-uv\fR). If not given, \fBuv\fR is looked up in \fBPATH\fR.
+
+.TP
+.BR --uv-pip-options =\fIUV_PIP_OPTION\fR
+Extra options to be supplied to \fBuv pip install\fR (this option can be used multiple times); only valid together with \fB--use-uv\fR.
```
facundobatista, line 158: Same detail for the text as I commented in argparse above.

tests/test_uvmanager.py
```diff
@@ -0,0 +1,125 @@
+# Copyright 2024-2026 Facundo Batista, Nicolás Demarchi
```
facundobatista, line 1: 2026 only

README.rst
```diff
@@ -391,6 +391,42 @@ Examples:
...
+
+``uv`` must be available: fades looks it up in ``PATH`` (installed via your system package manager, the standalone installer, etc.; fades does not depend on the ``uv`` PyPI package). You can point at a specific executable with ``--uv-path /path/to/uv`` (which also implies ``--use-uv``). If ``uv`` is requested but can't be found, fades errors out.
+
+Everything else stays fades: dependency parsing (``# fades`` comments, docstring, ``-d``/``-r``), venv indexing and reuse, ``--where``/``--rm``/``--clean-unused-venvs``, config files, the REPL and ``--autoimport``, etc.
+
+Some notes when using ``uv``:
+
+- fades runs ``uv venv --seed``, so the virtual environment ships ``pip`` (and ``setuptools``, depending on the Python version) just like the classic ``venv`` backend, keeping packages that expect them present working.
+- The PyPI availability pre-check is skipped (``uv`` resolves directly and fails early on its own).
+- ``--pip-options`` and ``--venv-options`` are **not** valid together with ``--use-uv`` (they target ``pip``/``venv``); use ``--uv-pip-options`` to pass extra options to ``uv pip install`` instead.
```
facundobatista, line 411: change "together with --use-uv" for something like "when uv indicated to use", aligned with argparse/man texts.

README.rst
```diff
@@ -391,6 +391,42 @@ Examples:
...
+
+Enabling uv permanently (via config)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+If you always want the uv backend, set ``use_uv=true`` under the ``[fades]`` section of any fades config file. fades reads (in increasing priority) ``/etc/fades/fades.ini`` (machine-wide), the per-user config (``<xdg-config>/fades/fades.ini``) and a repo-local ``./.fades.ini`` (handy to enable uv for a single project). A ``--use-uv``/``--no``-style flag on the command line still wins over the config.
```
facundobatista, line 416: No need to explain again how the config files work, just refer to "Setting options using config files" section.

README.rst
```diff
@@ -391,6 +391,42 @@ Examples:
...
+
+To turn it on for your user without editing files by hand, run::
+
+    python -W ignore -c "import configparser; from pathlib import Path; from fades.helpers import get_confdir; \
+    p = Path(get_confdir()) / 'fades.ini'; c = configparser.ConfigParser(); c.read(p); \
+    c.has_section('fades') or c.add_section('fades'); c.set('fades', 'use_uv', 'true'); \
+    c.write(p.open('w')); print('uv enabled by default in', p)"
```
facundobatista, line 423: I have mixed feelings about this code block...
- "editing files by hand" is a problem for a `fades` user?
- if we leave this recommendation on all config parameters, the README is a mess
- the code block is quite ugly (not sure that it can be improved without losing portability)
- maybe something like this corresponds to the "Setting options using config files" section?

fades/envbuilder.py
```diff
@@ -115,11 +117,49 @@ def post_setup(self, context):
...
+    except Exception as error:
+        logger.exception("Error creating virtual environment with uv:  %s", error)
+        raise FadesError("General error while creating venv with uv")
+
+    env_bin_path = helpers.get_env_bin_path(env_path)
+    return env_path, env_bin_path
+
+
+def create_venv(requested_deps, interpreter, is_current, options, pip_options, avoid_pip_upgrade,
+                use_uv=False, uv_exe=None, uv_pip_options=None):
```
facundobatista, line 148: I don't like this function's interface. It's weird (bunch of options together not working with other bunch of options). It's hard to check all combinations (e.g. having `uv_exe` if `use_uv`, but also that `avoid_pip_upgrade` is not needed, etc).

I suggest the following refactor:

- have two entry points: `create_venv_classic` and `create_venv_uv`
- the caller in `main.py` is responsible of deciding which to use
    - this will make the interface of each function simpler and will make sense
    - no need to weird verifications
- each function will do a little of work and will call a common one
    - this common one would be completely agnostic
    - specially if it receives a "manager" (UvManager or PipManager instance, according to the caller)

What do you think?

fades/helpers.py
```diff
@@ -88,6 +89,17 @@ def logged_exec(cmd):
...
 
+def get_uv_exe(uv_path=None):
+    """Return the path to the ``uv`` binary, or None if it can't be found.
+
+    If ``uv_path`` is given that exact location is validated (it must exist and be executable),
+    otherwise ``uv`` is looked up in PATH. fades runs against the system Python (it is not
+    installed inside a venv), so we rely on a ``uv`` *binary* rather than the ``uv`` PyPI package,
+    which would require polluting the system Python.
+    """
+    return shutil.which(uv_path) if uv_path else shutil.which("uv")
```
facundobatista, line 100: I wonder if we should log in DEBUG here, for eventually be sure which `uv` is being used...

facundobatista (changes requested):
Great work! I annotated some questions and details. Thanks!

PR #455, opened 2026-06-20
Allow --rm to remove a virtualenv by its dependencies

gilgamezh:
## What

`--rm` now accepts **either** a UUID **or** nothing. When given as a bare flag it removes the virtualenv matching the indicated dependencies, so this:

```bash
fades --rm $(fades --where -d django | awk -F "/" '{print $NF}')
```

becomes simply:

```bash
fades --rm -d django
```

## How

- `--rm` uses `nargs='?'` with a `REMOVE_BY_DEPS` sentinel as `const`, distinguishing "not passed" / "passed with UUID" / "passed bare".
- The removal block was moved below the dependency/interpreter/options setup. With the sentinel it resolves the venv via `get_venv(indicated_deps, interpreter, options=options)` — the same matching used to *run* a venv — otherwise it uses the existing UUID lookup. Both share the destroy + warning logic.
- Backward compatible: `fades --rm <uuid>` behaves exactly as before.

## Tests

There were no tests for the `--rm` flow in `go()`. Added `RemoveVenvTestCase` covering removal by UUID (found / not found / invalid env_path) and by dependencies (found / not found).

- Full suite: 296 passed
- flake8: clean

## Docs

Updated `man/fades.1` and `README.rst` (usage synopsis + command examples).

### DIFF
--- a/fades/main.py
+++ b/fades/main.py
@@ -45,6 +45,10 @@ from fades.logger import set_up as logger_set_up
 # the rest of the module just fine
 logger = logging.getLogger('fades')
 
+# sentinel used as the '--rm' value when it's given as a bare flag (no UUID), meaning
+# "remove the virtualenv that matches the indicated dependencies"
+REMOVE_BY_DEPS = object()
+
 # the signals to redirect to the child process (note: only these are
 # allowed in Windows, see 'signal' doc).
 REDIRECTED_SIGNALS = [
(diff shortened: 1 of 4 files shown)


fades/main.py
```diff
@@ -260,8 +264,10 @@ def go():
         '--python-options', action='append', default=[],
         help="extra options to be supplied to python (this option can be used multiple times)")
     parser.add_argument(
-        '--rm', dest='remove', metavar='UUID',
-        help="remove a virtualenv by UUID; see --where option to easily find out the UUID")
+        '--rm', dest='remove', metavar='UUID', nargs='?', const=REMOVE_BY_DEPS, default=None,
```
facundobatista, line 267: This `default=None` is not used, right? Or have a behaviour that escapes me?

man/fades.1
```diff
@@ -91,8 +91,8 @@ The dependencies can be indicated in multiple places (in the Python source file,
 Execute the \fIchild_program\fR in the context of the virtual environment. The child_program, which in this case becomes a mandatory parameter, can be just the executable name (relative to the venv's bin directory) or an absolute path.
 
 .TP
-.BR --rm " " \fIUUID\fR
-Remove a virtual environment by UUID.  See \fB--get-venv-dir\fR option to easily find out the UUID.
+.BR --rm " " [\fIUUID\fR]
+Remove a virtual environment.  If a UUID is given, the environment with that UUID is removed (see \fB--get-venv-dir\fR option to easily find out the UUID); if no UUID is given, the environment matching the indicated dependencies is removed (e.g. \fBfades --rm -d django\fR).
```
facundobatista, line 95: The message in ArgumentParser mentions `--where`, which I think it's better that `--get-venv-dir` (in any case, we should be consistent)

facundobatista (changes requested):
Awesome, thanks! There's a couple of very small details, commented inline...

